PR #11, opened 2023-11-27
Add GitHub Actions workflow for code checks

InensusDev:


dmohns (commented):
This file is in the wrong location. Can you move it to `.github/workflows/check.yml` so it runs on this Pull Request and we can see some results?

dmohns:
Right now we are adding a single job that runs **all** the linter. Could you split the `build` pipeline into multiple ones, one per linter?

For example, currently your CI yaml looks (a little simplified) like this:

```yaml
jobs:
  build:
    - Lint ESLint JS
    - Lint ESLint Vue
    - Lint PHP
    - Run`npm build`    
```

Could you change it to be more like this:

```yaml
jobs:
  lint-eslint-js:
    - Lint ESLint JS
  lint-eslint-vue:
    - Lint ESLint Vue
  lint-php:
    - Lint PHP
  build:
    - Run`npm build`    
```

This way they can all run in parallel and don't get blocked if one of them fails.

PR #14, opened 2023-12-01
(PhpStan) fix analyzer level = 1 issues

alchalade:
This MR fixes the issues which are listed by  phpstan.level 1.  The goal is reach level 8 but that would be a long journey. 

```
level1: possibly undefined variables, unknown magic methods and properties on classes with __call and __get
```
output for level 8
image

For more details abou the rule levels : https://phpstan.org/user-guide/rule-levels


dmohns:
Hey @alchalade if I understand correctly, after this PR, if we run `phpstan` with rule level 1 over the code base no warnings will be raised? If you, could you also add a small Github Actions that does exactly this?

dmohns:
> @dmohns I think this phpstan.level=1 check is already in a gitaction. Please check the **laravel-lint** action. Its one of the steps. For now it will catch us but won't block the PR.
> 
> After this one is merged, I will increase the level to 5 and try to address those issues. But for now level 1 should already run on new code and all reported issues should be addressed.

Ohh, I see. So, we "could" remove the `continue-on-error: true` from the PHP Lint action as part of this PR. But totally fine to do it in a different PR as well.

dmohns (dismissed):

dmohns:
Nice :shipit:

dmohns (approved):
Nice :shipit:

Issue #25, opened 2024-01-24
Composer throws `Class "SoapClient" not found` on initial run


dmohns:
## Description

On the initial run of `docker compose up` the Composer throws an error 

```
   Error 

  Class "SoapClient" not found

  at app/ManufacturerApi/CalinReadMeter.php:43
     39▕      */
     40▕     public function __construct(MeterConsumption $consumption)
     41▕     {
     42▕         try {
  ➜  43▕             $this->api = new SoapClient(config('services.calin.meter.api'), ['keep_alive' => false]);
     44▕         } catch (\Exception $exception) {
     45▕         }
     46▕         $this->consumption = $consumption;
     47▕     }
```

I'm not sure whether or not this is actually an issue, as in, negatively affects function of the app.

## Steps to re-produce

On a fresh install run `docker compose up` and check the log out from `mpmanager_composer` container.

PR #27, opened 2024-01-24
(infra) Remove composer container and let laravel container install d…

alchalade:
#25 

The laravel conitainer explicitly installs the soap extension in order to use the soapClient. After evaluating both
- "Creating a custom composer image "
-  "let the laravel container install the dependencies"

 I decided to go with the 2nd option as having a separate composer container has no extra value over let the main php container install the dependencies.

## How to test;
- remove **vendor** directory
- run `docker-compose up --build` this should already re-build the **laravel** container
- expect to see the **vendor** directory in `Website/htdocs/mpmanager/`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,14 +59,6 @@ services:
       - maria:db
       - redis:redis
 
-  composer:
-    container_name: mpmanager_composer
-    image: composer:2.6.5
-    restart: "no"
-    command: install --ignore-platform-reqs
-    volumes:
-      - ./Website/htdocs/mpmanager:/app
-
   redis:
     image: redis:5
     volumes:
--- a/Docker/DockerfileLaravelDev
+++ b/Docker/DockerfileLaravelDev
@@ -16,14 +16,12 @@ RUN apt-get update && \
     git \
     zsh
 
-
 # install ohmyzsh
 RUN chsh -s $(which zsh)
 RUN sh -c "$(curl -fsSL https://raw.github.com/ohmyzsh/ohmyzsh/master/tools/install.sh)"
 
 # add php artisan alias
 RUN echo 'alias "cmd=php artisan"'  >> ~/.zshrc
-
 
 # remove apt lists
 RUN rm -rf /var/lib/apt/lists/*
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -17,6 +17,8 @@ services:
     volumes:
       - ./Website/config/php/php.ini:/usr/local/etc/php/php.ini
       - ./Website/htdocs:/var/www/html
+    command: [ "sh", "-c", "cd /var/www/html/mpmanager && php composer.phar install" ]
```
dmohns, line 20: Doesn't this overwrite the `CMD` from the underlying Dockerfile `CMD ["apache2-foreground"]` ?https://hub.docker.com/layers/library/php/8.2-apache/images/sha256-39d43cc32ea91e96fa7333cd52f67b6d91c8c4b22130443db0b1ac5f26b7c738?context=explore ?

alchalade:
@dmohns oh very good catch thank you for this 🙏 . I totally missed that one.

docker-compose.yml
alchalade: @dmohns oh very good catch thank you for this 🙏 . I totally missed that one.

dmohns:
I'm trying to run this on my local machine after deleting the `vendor` directory.

Running `docker compose up` does indeed re-create the `vendor` directory and I can see output of composer in the logs. However, the container exists and doesn't keep running. This isn't intended right?

alchalade:
@dmohns TBH I didn't checked it before I pushed the 2nd change. The weird thing is that the CMD form the base image got lost even though there is nothing in between that overwrites the `CMD` .

The defensive check in the .sh file should do the trick now.

dmohns:
> @dmohns TBH I didn't checked it before I pushed the 2nd change. The weird thing is that the CMD form the base image got lost even though there is nothing in between that overwrites the `CMD` .
> 
> The defensive check in the .sh file should do the trick now.

Yepp, it works now. I will admit I don't fully understand why this happens, but that's fine 😉 

LGTM

dmohns (dismissed):

dmohns:
Little Github hack: You can use 

```
Closes #25
```

syntax for this to automatically close the linked issue, once the PR is merged 😉

PR #41, opened 2024-02-21
Run local development environment without `hosts` file modifications

dmohns:
The goal of this PR is to further simplify the setup of a local development environment. 

Currently, the local development environment uses a dedicated Nginx container to route local traffic. For this to work developers are required to modify their hosts file. 

This is cumbersome for multiple reasons

- Errors when modifying the hosts file can cause system wide issues
- It might interfere with other projects the developer is trying to run


dmohns:
> If a developer struggles to add a hosts entry it's time to think alternative career paths. 

Hehe 😉 

> I don't see any value by replacing the named local domain with local host.

FWIW it's one more step a developer needs to do to run the project. 

Actually, it's the other way around: What is the value of using local domain that justifies an additional step? If there is any value then I'm fine of having this additional setup step. But I'm struggling to see the benefit.

PR #49, opened 2024-02-27
MPM-9 Add `usageType` to main settings view

dmohns:
This is a first step towards making the Usage Type customisable. This PR adds the possibility for users to select `usageType` in the main settings dialogue. It's not integrated into the create customer flow yet.

**Preview:**
image

Based on the selection on `usageType`, the plugin view will be adapted and in-compatible plugins will be high lighted

**Preview (for Mini-Grid customer):**
image

Plugins that do not fit to usage types are hidden.

**Preview (for a E-Bike customer):**
image


InensusDev (commented):
Description

We require modifications to the company registration stepper process.

In Step 1 (Company form), we will introduce a new dropdown menu that presents a list of usage_types.

In Step 2 (Plugin selection), the plugin list displayed will be determined by the usage_type selected in Step 1. Users will only be able to select plugins appropriate for the chosen usage_type. Plugins will fall into one of the following categories: 'e-bike', 'shs', 'mini-grid', or 'gen'. Additionally, plugins categorized as 'gen' will be displayed regardless of the usage_type selected.

Website/ui/src/modules/Settings/PluginSettings.vue
```diff
@@ -32,6 +34,21 @@ export default {
...
             type: Array,
             required: true
+        },
+        mainSettings: {
+            type: Object,
+            required: true
+        }
+    },
+    computed: {
+        enriched_plugins: function() {
```
InensusDev, line 44: Let's use camelCase every where we can. There could be forgotten samples as snake_case, It would be great if we could refactor them.

dmohns: Done, as far as the code changed in this PR is concerned.

I like the idea of changing it everywhere. But let's do that in a separate change.

PR #116, opened 2024-06-24
deploy/deployment

Delyc:
Frontend deployment


composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "tymon/jwt-auth": "^2.1"
```
musayann, line 3: delete this file as well. I don't think we need it

Docker/DockerfileLaravelDev
```diff
@@ -1,49 +1,24 @@
-FROM php:8.2-apache
+FROM php:8.2-fpm-alpine
```
musayann, line 1: reset changes in this file. I think we are not using them

composer.lock
```diff
@@ -0,0 +1,4861 @@
+{
+    "_readme": [
```
musayann, line 2: delete this file as well. we don't need it

musayann (changes requested):

Issue #122, opened 2024-06-26
Route and Sidebar Management Clean-up


dmohns:
### Scope:

Currently, route and sidebar management involves a complex interaction between the backend and frontend components:

1. The backend writes data into the menu_items and sub_menu_items tables.
2. The frontend applies implicit and hidden filtering via ExportedRoutes and transforms this data for the sidebar. This process is detailed in the codebase here: [SideBar.vue](https://github.com/EnAccess/micropowermanager/blob/main/Website/ui/src/modules/Sidebar/SideBar.vue#L148-L164).

This approach was adopted in the legacy MPMv1 to facilitate external backend plugins in registering their UI components easily. However, the current version of MPM does not support external plugins, as all provider packages are integrated directly into the application.

The existing method now introduces unnecessary complexity and should be simplified to enhance maintainability.

### Definition of done:

-  **Single Source of Truth:** Centralize all route management in a single source, likely `ExportedRoutes.js` via the `meta` data similar to how Breadcrumb navigation is achieved. If an alternative location is deemed more appropriate, document the rationale for the chosen place.

- **No Redundant Code:** Eliminate the backend logic associated with menu_items and sub_menu_items tables.
 Update Frontend Logic: Refactor or remove the current filtering and transformation processes in the frontend to streamline route management.
 
- **Up to date Developer Documentation:** Update the developer documentation to reflect these changes, ensuring it includes detailed instructions on the new route and sidebar management process.

PR #132, opened 2024-07-01
Render sidebar from routes

dmohns:
Makes progress on https://github.com/EnAccess/micropowermanager/issues/122

The goal of this PR is to render the Sidebar from `routes` definition via newly added meta arguments.

The advantage is that the entire navigation is managed in one place. Rather than a hybrid of Frontend routes beging defined in `routes.js` and then wrapped in backend tables.


dmohns:
Continued in: https://github.com/EnAccess/micropowermanager/pull/141

Issue #110, opened 2024-06-18
Update composer to latest version


dmohns:
The repository ships with a very old version of composer (which in fact is affected by a CVE).

```sh
>> php ../composer.phar --version
Composer version 2.0.4 2020-10-30 22:39:11
```

Please upgrade to the latest version of composer.

This is blocked by: https://github.com/EnAccess/micropowermanager/issues/96

Issue #102, opened 2024-06-17
Fix `...does not comply with psr-4 autoloading standard.`


dmohns:
After `composer install` we encounter a lot `... does not comply with psr-4 autoloading standard.` errors. Please fix these

```
laravel     | Class Tests\Feature\UserResource located in ./tests/Feature/UserResourceTest.php does not comply with psr-4 autoloading standard. Skipping.
laravel     | Class Inensus\SwiftaPaymentProvider\Tests\Unit\PaymentProcessTests located in ./packages/inensus/swifta-payment-provider/src/tests/Unit/PaymentProcessTests.php does not comply with psr-4 autoloading standard. Skipping.
laravel     | Class Inensus\SwiftaPaymentProvider\Http\Exceptions\CustomerAddressNotFoundException located in ./packages/inensus/swifta-payment-provider/src/Exceptions/CustomerAddressNotFoundException.php does not comply with psr-4 autoloading standard. Skipping.
laravel     | Class Inensus\SwiftaPaymentProvider\Http\Exceptions\SwiftaValidationFailedException located in ./packages/inensus/swifta-payment-provider/src/Exceptions/SwiftaValidationFailedException.php does not comply with psr-4 autoloading standard. Skipping.
laravel     | Class Inensus\SwiftaPaymentProvider\Http\Exceptions\CipherNotValidException located in ./packages/inensus/swifta-payment-provider/src/Exceptions/CipherNotValidException.php does not comply with psr-4 autoloading standard. Skipping.
laravel     | Class Inensus\SwiftaPaymentProvider\Http\Exceptions\TransactionAmountDifferentException located in ./packages/inensus/swifta-payment-provider/src/Exceptions/TransactionAmountDifferentException.php does not comply with psr-4 autoloading standard. Skipping.
laravel     | Class Inensus\SwiftaPaymentProvider\Http\Exceptions\TransactionNotExistsException located in ./packages/inensus/swifta-payment-provider/src/Exceptions/TransactionNotExistsException.php does not comply with psr-4 autoloading standard. Skipping.
laravel     | Class Inensus\StronMeter\Tests\Unit\TokenGenerationTest located in ./packages/inensus/stron-meter/src/tests/Unit/TokenGenerationTest.php does not comply with psr-4 autoloading standard. Skipping.
[...]

Issue #96, opened 2024-06-12
Make code run with Composer 2.7


dmohns:
Somehow our code base doesn't run with composer 2.7

Please execute all the changes that make it run with latest version of Composer.

**Further information:**

- Temp downgrade: https://github.com/EnAccess/micropowermanager/pull/95
- Common upgrade problems: https://github.com/composer/composer/releases/tag/2.7.0
- Example or error: https://github.com/EnAccess/micropowermanager/actions/runs/9485629323/job/26138019558

PR #249, opened 2024-08-10
Remove composer.phar from file system

alchalade:
With this MR;
-  the composer.phar won't be shipped a as file from VCS. Instead Docker will pull the latest version from getcomposer and will make it available in the `laravel` container.
-  it also addresses the psr-4 warnings
- drops the outdated phpMyAdmin database management tool. There are plenty moderns database tools like table plus, mysql workbench the consumers/developers should consider to use one of them 

Closes: #96
Closes: #102
Closes: #110


Docker/entrypoint.sh
```diff
@@ -1,6 +1,15 @@
...
+#download composer
+php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+php -r "if (hash_file('sha384', 'composer-setup.php') === 'dac665fdc30fdd8ec78b38b9800061b4150413ff2e3b6f88543c636f7cd84f6db9189d43a81e5503cda447da73c7e5b6') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
+php composer-setup.php
+ls -la
+php -r "unlink('composer-setup.php');"
+mv composer.phar /usr/local/bin/composer
+
+composer install
+
```
dmohns, line 12: Could you please move this entire section into the Docker build file itself? I didn't realise we are running `composer install` from the entrypoint script. It should be part of the immutable image rather than the runtime container.

dmohns (commented):
Thanks for the PR 🥳 Looks good, just a few general comments and one comment in the code

1. Could you please make sure your commit email address is linked to your Github account?
2. Could you link the issues this PR is resolving in the comment (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) for example `Closes #96`

Docker/entrypoint.sh
alchalade: Its tricky to move the entrypoint to the dockerfile. As far as I know, the entry point is triggered right after the container creation. As the current state mounts the volume while the container creation `composer install` should be done when the files are present. 

We could potentially move Line:3 to 10 into the docker-file but having them here makes contextually more sense

dmohns: Alright, I see... we need this to be done on the runtime side to get hot-reloads in the dev server.

So let's leave it like it is then. However, could you make it more clear that the `entrypoint` script is specific to the dev container? For example by re-naming it to `start_dev_container` or something like that.

dmohns (approved):

PR #274, opened 2024-09-18
feat: built frontend and backend production docker image

Delyc:
Building Frontend and backend production docker image

### DIFF
--- a/Docker/DockerfileLaravelProd
+++ b/Docker/DockerfileLaravelProd
@@ -1,4 +1,7 @@
-FROM php:8-fpm
+FROM php:8.2-apache
+
+WORKDIR /app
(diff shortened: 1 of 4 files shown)


Docker/DockerfileLaravelProd
```diff
@@ -1,4 +1,7 @@
-FROM php:8-fpm
+FROM php:8.2-apache
```
dmohns, line 1:
```suggestion
FROM php:8.2-fpm
```

I think we should prefer to keep using `fpm`. https://www.php.net/manual/en/install.fpm.php

dmohns (approved):
LGTM :shipit:

PR #275, opened 2024-09-19
Ft/production dockerfile

Delyc:
This PR moves the frontend production docker file from `Website/UI` directory to `Docker` directory


dmohns:
Hey @Delyc you are pulling in commit's that are already merged. Can you please rebase your branch to current `main` ?

dmohns:
Two more comments:

1. You are not moving the file, but copying it. Could you make sure to move it, i.e. so we don't have the file twice.
2. Also, it would be great if we can move the Dev file for UI. But would probably require to change the Docker Compose file

dmohns (approved):

musayann (approved):

dmohns:
Hey @Delyc 

This PR got a bit messy. Github doesn't detect the Dockerfile was moved, which would be great for organisational purposes. 

I'm closing this PR. Could you please re-create this PR with 

- a single commit that does only the move
- Additional commits that do the changes

?

For example: https://github.com/EnAccess/micropowermanager/pull/279/files

PR #276, opened 2024-09-22
fix steamaco meter sync process

### LINKED ISSUE #183: `SteamaCo` plugin: Broken call to `associate` Tariff on Meter sync
The Meter sync feature of the `SteamaCo` plugin calls `associate` on MPM Tariff. However, it never actually "creates" the Tariff in the database.

It's unclear whether this is 

1. User responsibility to provide the Tariffs according to Meter configuration in SteamaCo
2. A bug in the code

alchalade:
this MR fixes the meter synchronization process of the SteamaCo plugin.

closes #183

### DIFF
--- a/Website/htdocs/mpmanager/app/Services/PluginsService.php
+++ b/Website/htdocs/mpmanager/app/Services/PluginsService.php
@@ -3,28 +3,22 @@
 namespace App\Services;
 
 use App\Models\Plugins;
-use App\Services\Interfaces\IBaseService;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Pagination\LengthAwarePaginator;
 
-/**
- * @implements IBaseService<Plugins>
- */
-class PluginsService implements IBaseService
+class PluginsService
 {
     public function __construct(
         private Plugins $plugin
     ) {
     }
 
-    public function getById(int $id): Plugins
-    {
-        throw new \Exception('Method getById() not yet implemented.');
-    }
-
     public function create(array $pluginData): Plugins
     {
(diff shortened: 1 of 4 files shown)


Website/htdocs/mpmanager/app/Services/PluginsService.php
```diff
@@ -53,4 +43,12 @@ public function getByMpmPluginId($mpmPluginId)
     {
         return $this->plugin->newQuery()->where('mpm_plugin_id', $mpmPluginId)->first();
     }
+
+    public function isPluginActive(int $SPARK_METER): bool
+    {
+        return $this->plugin->newQuery()
+            ->where('mpm_plugin_id', '=', $SPARK_METER)
+            ->exists();
+    }
```
dmohns, line 52: Can you rename the function argument? `$SPARK_METER` is misleading, as this can be called with any `plugin_id`, right?

dmohns (commented):
Hey @alchalade 

looks good, thanks for the work here. Two main comments:

1. Could you check the failing CI?
2. Could you quickly explain what was the problem here? Most of this PR's changes seem like simple refactoring (which I like a lot), but that don't change the logic. Which of the changes is the actual bug fix?

Also, minor comment inside the code.

Website/htdocs/mpmanager/app/Services/PluginsService.php
alchalade: sure, the function implementation was created by the IDE based on the call from the spark meter package.

dmohns (approved):
:shipit:

Issue #100, opened 2024-06-17
Explore Laravel Pint as a linter


dmohns:
Being a Laravel project maybe it would be great to also adhere to the "official" Laravel Code Style: https://laravel.com/docs/11.x/pint

pechtelt1:
@dmohns i think this is a good idea. Do you have some code style requirements or do you just want to have the default Laravel Pint?

munyanezaarmel:
@dmohns If this issue is still relevant, I would be happy to take it on.

dmohns:
Sure, it's still somewhat relevant. Feel free to take a look 👍 

Bear in mind that we recently adopted PHP-CS-Fixer, see https://github.com/EnAccess/micropowermanager/pull/99 So we'd have to check how the two interact.

dmohns:
We are not going to use **Laravel Pint** as we already have PHP-CS-Fixer. For more details, see response here: https://github.com/EnAccess/micropowermanager/pull/364#issuecomment-2500511460

PR #364, opened 2024-11-20
Laravel Pint as a linter #100

munyanezaarmel:

closes:#100 
### Pull Request checklist


.github/workflows/lint.yml
```diff
@@ -0,0 +1,35 @@
...
+        working-directory: ./src/backend
+ 
+      - name: Run Pint
+        run: pint
+        working-directory: ./src/backend
+ 
+      - name: Commit linted files
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: "Fixes coding style"
```
dmohns, line 35: I know this is the way that is given in the Pint documentation, but I'm not a huge fan of this. Could we instead only run `pint` is check mode? I.e. just have the CI fail, and leave it to developers to actuallly fix the lint errors?

munyanezaarmel:
@dmohns can you please check this

dmohns:
> @dmohns can you please check this

Sure, I will follow up with this next week 👌

dmohns:
Hey @munyanezaarmel 

apologies for my delayed response here. I will take a look now 🕵️

dmohns:
Hey @munyanezaarmel 

so, after reviewing this here are my thoughts. It looks like **Laravel Pint** is actually just a rule-set to PHP-CS-Fixer plus a small wrapper around it for nicer CLI integration. The exact rule configuration can be seen here: https://github.com/laravel/pint/blob/main/resources/presets/laravel.php

Now, I don't think there is a benefit of adding the tool it self then. I was originally hoping there is some additional rules and check in there, which PHP-CS-Fixer cannot cover. But this doesn't seem to be the case.

None-the-less... I think some of the conventions that the Laravel rule set I actually like a lot. So there is something to take a wat from here.

Could you instead please adapt our current PHP-CS-Fixer rule set with the following rules:

```php
        'method_chaining_indentation' => true,
        'braces_position' => [
            'control_structures_opening_brace' => 'same_line',
            'functions_opening_brace' => 'same_line',
            'anonymous_functions_opening_brace' => 'same_line',
            'classes_opening_brace' => 'same_line',
            'anonymous_classes_opening_brace' => 'same_line',
            'allow_single_line_empty_anonymous_classes' => true,
            'allow_single_line_anonymous_functions' => true,
        ],
        'method_argument_space' => [
            'on_multiline' => 'ensure_fully_multiline',
        ],
        'single_line_empty_body' => true,
```

?

dmohns:
Hey @munyanezaarmel 

for changing the rules set of PHP-CS-Fixer, could you do the following:

1. Create a new branch (please dont use special characters like `#` in branch names)
2. Create exactly 2 commits in the branch. First is to update the rule set
3. Second commit should reformat the entire repo

Then we will merge this PR. 

4. After that please create a new PR to add the commit that reformatted the repo to `.git-blame-ignore-revs`

Issue #396, opened 2024-12-05
[Feature Request]: Display Tokens in the application


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Currently, there isn't really a way to show Energy Tokens in the app. The only way is via checking the messages sent to a customer.

### Proposed Solution

In the Meter Transaction view:

![Image](https://github.com/user-attachments/assets/1c611b62-fb5e-4c1d-9c57-e7997c388710)

It would be great if we can somehow see the Token. Either via making the list clickable. Or show the Token in the same list.

### Alternatives Considered

N/A

Issue #83, opened 2024-05-30
Token Separation


dmohns:
The tokens for the Angaza units come separated which helps in readability, can the same to done for the Sunking ones? (eg. `123 231 112 111` instead of `123231112111`);

PR #400, opened 2024-12-08
use correct paid_for_type on payment #396

beesaferoot:
- seed meter tokens table for demo company
- use appropriate paid_for_type as expected on the frontend

This PR seeds the meter_tokens table, which ensures the `tokens` field is correctly returned with values on the `api/meters/{meter_id}/` endpoint.

Also on success payment, the payment history `paid_for_type` is updated to match expected values on the frontend. 

Closes: #396
Closes: #83

We can now see the tokens on the transaction 

Screenshot 2024-12-08 at 8 55 17 PM

### Pull Request checklist

### DIFF
--- a/src/backend/app/Console/Commands/DemoDataCreator.php
+++ b/src/backend/app/Console/Commands/DemoDataCreator.php
@@ -2,9 +2,11 @@
 
 namespace App\Console\Commands;
 
+use App\Helpers\TokenGenerator;
 use App\Models\MainSettings;
 use App\Models\MaintenanceUsers;
 use App\Models\Meter\Meter;
+use App\Models\Meter\MeterToken;
 use App\Models\Person\Person;
 use App\Models\Token;
 use App\Models\Transaction\AgentTransaction;
@@ -47,6 +49,7 @@ class DemoDataCreator extends AbstractSharedCommand {
         private AirtelTransaction $airtelTransaction,
         private Meter $meter,
         private Token $token,
+        private MeterToken $meterToken,
         private CalinTransaction $calinTransaction,
         private MainSettings $mainSettings,
         private TicketCategory $ticketCategory,
@@ -254,7 +257,7 @@ class DemoDataCreator extends AbstractSharedCommand {
         // generate random token
         if ($transactionData->transaction->amount > 0) {
             $tokenData = [
-                'token' => Str::random(30),
+                'token' => TokenGenerator::generate(),
                 'load' => round(
                     $transactionData->transaction->amount /
                         $randomMeter['tariff']['price'],
@@ -265,6 +268,22 @@ class DemoDataCreator extends AbstractSharedCommand {
             $token->transaction()->associate($transaction);
             $token->save();
             $transactionData->token = $token;
+
+            // generate meter_token
+            $meterTokenData = [
+                'meter_id' => $randomMeter->id,
+                'token' => TokenGenerator::generate(),
+                'energy' => round(
+                    $transactionData->transaction->amount /
+                    $randomMeter['tariff']['price'],
+                    2
+                ),
+                'transaction_id' => $transaction->id,
+            ];
+            $meterToken = $this->meterToken->newQuery()->make(['meter_id' => $meterTokenData['meter_id'],
+                'token' => $meterTokenData['token'], '' => $meterTokenData['energy'],
+                'transaction_id' => $meterTokenData['transaction_id']]);
+            $meterToken->save();
 
             // payment event
             event(
--- a/src/backend/app/Helpers/TokenGenerator.php
+++ b/src/backend/app/Helpers/TokenGenerator.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace App\Helpers;
+
+class TokenGenerator {
+    /**
+     * Generate a random 12-digit token.
+     *
+     * @return string
+     */
+    public static function generate() {
+        $token = '';
+        for ($i = 0; $i < 12; ++$i) {
+            $token .= random_int(0, 9);
+        }
+
+        return $token;
+    }
+}
--- a/src/frontend/src/modules/Meter/Transactions.vue
+++ b/src/frontend/src/modules/Meter/Transactions.vue
@@ -94,10 +94,10 @@ export default {
     },
     formatToken(token) {
       // Ensure token is a string
-      const tokenStr = String(token);
+      const tokenStr = String(token)
       // Format in the desired pattern
       // return tokenStr.match(/.{1,4}/g).join('-'); // For "1234-1234-1234"
-      return tokenStr.match(/.{1,3}/g).join(' '); // For "123 412 341 234"
+      return tokenStr.match(/.{1,3}/g).join(" ") // For "123 412 341 234"
     },
   },
 }
--- a/src/frontend/src/modules/Meter/Transactions.vue
+++ b/src/frontend/src/modules/Meter/Transactions.vue
@@ -30,8 +30,8 @@
               {{ token.paid_for_type }}
             </md-table-cell>
             <md-table-cell
-              v-if="token.paid_for_type === 'token'"
-              v-text="readable(token.paid_for.energy) + 'kWh'"
+              v-if="token.paid_for_type === 'App\\Models\\Token'"
+              v-text="readable(token.paid_for.energy) + ' kWh'"
             ></md-table-cell>
             <md-table-cell v-else>-</md-table-cell>
             <md-table-cell
@@ -49,10 +49,11 @@ import Widget from "../../shared/widget"
 import { EventBus } from "@/shared/eventbus"
 import { currency } from "@/mixins/currency"
 import { timing } from "@/mixins/timing"
+import { token } from "@/mixins/token"
 
 export default {
   name: "Transactions.vue",
-  mixins: [currency, timing],
+  mixins: [currency, timing, token],
   components: { Widget },
   props: {
     transactions: {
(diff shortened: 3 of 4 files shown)


src/backend/app/Listeners/PaymentListener.php
```diff
@@ -86,11 +86,11 @@ public function onPaymentSuccess(
                 $this->applianceRatePaymentHistoryService->assign();
                 break;
             case $paidFor instanceof Asset:
-                $paymentHistory->paid_for_type = Asset::class;
+                $paymentHistory->paid_for_type = 'appliance';
```
dmohns, line 89: This looks like a massive change. Probably it's correct, but I'm curious: How did you test this?

src/backend/app/Console/Commands/DemoDataCreator.php
```diff
@@ -265,6 +267,22 @@ private function generateTransaction(): void {
...
+                    $transactionData->transaction->amount /
+                    $randomMeter['tariff']['price'],
+                    2
+                ),
+                'transaction_id' => $transaction->id,
+            ];
+            $meterToken = $this->meterToken->newQuery()->make(['meter_id' => $meterTokenData['meter_id'],
+                'token' => $meterTokenData['token'], '' => $meterTokenData['energy'],
+                'transaction_id' => $meterTokenData['transaction_id']]);
+            $meterToken->save();
```
dmohns, line 284: Did you manage to understand what is the difference between `meterToken` and the `token` from above?

src/backend/app/Console/Commands/DemoDataCreator.php
```diff
@@ -265,6 +267,22 @@ private function generateTransaction(): void {
             $token->save();
             $transactionData->token = $token;
 
+            // generate meter_token
+            $meterTokenData = [
+                'meter_id' => $randomMeter->id,
+                'token' => Str::random(30),
```
dmohns, line 273: Another nit here: Instead of random 30-char strings, can we can 12 digit numbers? `[0-9]` ? (Probably also applies to the code piece above)

src/backend/app/Listeners/PaymentListener.php
beesaferoot: I checked on the frontend, the Meter Transanction list has a condition to display approriate message based on the paid for type. 


```js
     <md-table-row v-for="token in transactions.tokens" :key="token.id">
            <md-table-cell
              v-text="token.transaction.original_transaction_type"
            ></md-table-cell>
            <md-table-cell
              v-text="moneyFormat(token.transaction.amount)"
            ></md-table-cell>
            <md-table-cell v-if="token.paid_for_type === 'token'">
              Token {{ token.paid_for.token }}
            </md-table-cell>
            <md-table-cell v-else>
              {{ token.paid_for_type }}
            </md-table-cell>
            <md-table-cell
              v-if="token.paid_for_type === 'token'"
              v-text="readable(token.paid_for.energy) + 'kWh'"
            ></md-table-cell>
            <md-table-cell v-else>-</md-table-cell>
            <md-table-cell
              v-text="timeForTimeZone(token.created_at)"
            ></md-table-cell>
          </md-table-row>
  ```
  
  However, on the backend, the string is updated to the internal class name (i.e `App\Model\Token`), which did not match with what the frontend expected; I found that weird. 
  
  Also, while searching for the usecases for `paid_for_type` field I found this factory. The values seem to tally with the frontend expectation
  
  ```php
  class PaymentHistoryFactory extends Factory {
    protected $model = PaymentHistory::class;

    /**
     * Define the model's default state.
     *
     * @return array
     */
    public function definition() {
        return [
            'id' => 1,
            'transaction_id' => 1,
            'amount' => $this->faker->randomFloat(2, 0, 100),
            'payment_service' => $this->faker->randomElement(['vodacom_transaction', 'airtel_transaction', 'agent_transaction']),
            'sender' => $this->faker->phoneNumber,
            'payment_type' => $this->faker->randomElement(['appliance', 'energy', 'installment', 'access rate']),
            'paid_for_type' => $this->faker->randomElement(['appliance', 'token', 'loan_rate', 'access_rate']),
            'paid_for_id' => 1,
            'payer_type' => 'person',
            'payer_id' => 1,
        ];
    }
}
```

src/backend/app/Console/Commands/DemoDataCreator.php
beesaferoot: Should this be a random 12 digits number or we want to have a special token generator?

src/backend/app/Listeners/PaymentListener.php
beesaferoot: What are your thoughts? Do you think we should have a different display for tokens? Right now the frontend also gets the token data we could potentially use for that (from the Meter Token table)

src/backend/app/Console/Commands/DemoDataCreator.php
dmohns: I think for now a simple 12 random digits is fine 👍

src/backend/app/Console/Commands/DemoDataCreator.php
dmohns: Basically, I'm looking at this change: https://github.com/EnAccess/micropowermanager/commit/c85f9aa0c6cdedb468dbe91a96f2c89448faa853#diff-588c01299a5ad9020a72f27600dc3dfd9cc8c4fbd8afaeb2917d0e9090f71d6d

What I'm wondering is: Is `MeterToken` deprecated? I.e. was `MeterToken` replaced by the more generic `Token` model?

src/backend/app/Listeners/PaymentListener.php
dmohns: There are two questions here. Let me address them individually:

**What should be the value for `paid_for`?** `App\Model\Token` vs `token`

If I understand correctly `paid_for` defines a [Polymorphic relation ship](https://laravel.com/docs/11.x/eloquent-relationships#one-of-many-polymorphic-relations). I would like to follow the practise that is standard in Laravel. On Database level it I think it should be the class type `App\Model\Token` as per Laravel convention of defining polymorphic relations (but I'm also not a Laravel expert, maybe I got this wrong), rather than the display name `token`.

So, I'd suggest we adapt the frontend filter to check for `App\Model\Token`.

**How should token be displayed?**

My preferred solution would be to adapt the Frontend to display Token objects like this:

`Token (1234-1234-1234)` or even `Token (123 412 341 234)` (this would also resolve https://github.com/EnAccess/micropowermanager/issues/83)

But this should be a pure frontend change. I.e. it should change the tokens in the backend, but just the way they are presented to the user

WDYT?

src/backend/app/Console/Commands/DemoDataCreator.php
beesaferoot: That sounds more plausible since the frontend does expect to get tokens from transactions. 

Alright, I will go with this assumption.

src/backend/app/Listeners/PaymentListener.php
beesaferoot: After reading up on how Laravel handles Polymorphic relationship, I now understand your explanations.  

I will revert this and make the frontend change as suggested. Thank you @dmohns

beesaferoot:
@dmohns Updated, kindly review.

src/frontend/src/modules/Meter/Transactions.vue
```diff
@@ -23,8 +23,8 @@
             <md-table-cell
               v-text="moneyFormat(token.transaction.amount)"
             ></md-table-cell>
-            <md-table-cell v-if="token.paid_for_type === 'token'">
-              Token {{ token.paid_for.token }}
+            <md-table-cell v-if="token.paid_for_type === 'App\\Models\\Token'">
+              Token ({{ formatToken(token.paid_for.token) }})
```
dmohns, line 27: We also need this 5 lines blow where it checks for the energy

src/frontend/src/modules/Meter/Transactions.vue
```diff
@@ -92,6 +92,13 @@ export default {
         this.transactions.tokens.length,
       )
     },
+    formatToken(token) {
+      // Ensure token is a string
+      const tokenStr = String(token)
+      // Format in the desired pattern
+      // return tokenStr.match(/.{1,4}/g).join('-'); // For "1234-1234-1234"
+      return tokenStr.match(/.{1,3}/g).join(" ") // For "123 412 341 234"
+    },
```
dmohns, line 101: Maybe you can move this into a JS mixing, similar to `currency` and `timing ` where formatting functions like `readable` (for energy) and `timeForTimeZone ` (for timing) are defined?

dmohns (commented):
Few small comments, other than that LGTM

beesaferoot:
Make some updates, please have a look @dmohns

src/backend/app/Helpers/TokenGenerator.php
```diff
@@ -0,0 +1,19 @@
...
+     */
+    public static function generate() {
+        $token = '';
+        for ($i = 0; $i < 12; ++$i) {
+            $token .= random_int(0, 9);
+        }
+
+        return $token;
+    }
+}
```
dmohns, line 19: We'll leave this for now, but one comment for a later improvement: This should probably be a Factory https://laravel.com/docs/11.x/eloquent-factories

dmohns (approved):
Thanks 🙏

dmohns:
One more non-code related comment: Please use [Pull Request keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) to link PR's to corresponding issues.

For example in this PR (in the description) you can do

```
Closes: #396
Closes: #83
```

beesaferoot:
@dmohns  Noted your feedbacks with thanks 🙏

PR #405, opened 2024-12-11
eks release workflow

Delyc:
### Pull Request checklist


.github/workflows/release-publish-cluster.yaml
```diff
@@ -0,0 +1,47 @@
+name: Deploy to EKS
+
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch:
```
dmohns, line 7: Having this as a separate workflow won't work. If a new release gets created this will run immediately. However, the deploy to DockerHub job takes around 1h usually. So, we Pull the Docker image before the new one is even available.

Could you instead, move the `deploy` job into the already existing `release-publish-dockerhub.yaml`, also rename it to `deploy-demo`. Then please add a `need` (see [here](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds)) to depend on the `release-publish`

dmohns (commented):
Left a few comments if you could address them

dmohns (approved):
Thanks!

PR #426, opened 2024-12-15
Fix test suite #108

### LINKED ISSUE #108: Fix the Test Suite
Currently, the test suite is broken. Please make it so that the `test` command works

```sh
root@4b3bb5556234:/var/www/html/mpmanager# php artisan test

   FAIL  Tests\Unit\AgentSellApplianceTest
  ⨯ agent sell appliance

   PHPUnit\Framework\ExceptionWrapper

  Database connection [shard] not configured.

  at vendor/laravel/framework/src/Illuminate/Database/DatabaseManager.php:173
    169▕         // If the configuration doesn't exist, we'll throw an exception and bail.
    170▕         $connections = $this->app['config']['database.connections'];
    171▕
    172▕         if (is_null($config = Arr::get($connections, $name))) {
  ➜ 173▕             throw new InvalidArgumentException("Database connection [{$name}] not configured.");
    174▕         }
    175▕
    176▕         return (new ConfigurationUrlParser)
    177▕                     ->parseConfiguration($config);


  Tests:  1 failed, 164 pending
  Time:   0.27s
```

beesaferoot:

This PR fixes #108 and creates a github test suite workflow

### Pull Request checklist


dmohns:
Looks like something isn't really working here. All tests are failing with 

```
SQLSTATE[HY000] [1045] Access denied for user 'forge'@'172.18.0.1' (using password: NO)
```

This suggests that it's not picking up the env variables. Can you check this?

Also, could you add instructions to the docs on how to run these tests locally, when developing the app?

src/backend/config/database.php
```diff
@@ -66,16 +66,38 @@
...
+            'port' => env('DB_PORT', '3306'),
+            'database' => env('DB_DATABASE', 'testing_test_company_db'),
+            'username' => env('DB_USERNAME', 'root'),
+            'password' => env('DB_PASSWORD', 'wF9zLp2qRxaS2e'),
+            'charset' => 'utf8mb4',
+            'collation' => 'utf8mb4_unicode_ci',
             'prefix' => '',
+            'strict' => true,
+            'engine' => null,
         ],
```
dmohns, line 83: I don't like that the tests use their own **connection** (own database yes, but not own connection). We actually have the connection names `micro_power_manager` and `shard` hardcoded in a lot of places in the code base. So, I'm doubtful if using different connection names will even work.

Do you think it's possible to remove the two testing connections and the envs `DB_CONNECTION` and `TEST_DB_CONNECTION`? 

We could instead use a different `DB_DATABASE` env when running test.

However, I'm not sure how many side effects this will have. So, I'd say we move this to a different PR

dmohns (approved):
One comment, but it should be a different PR, so I'm approving this one 😉

src/backend/config/database.php
beesaferoot: I will have to give it a try. I will make things cleaner if we have the same connections

Issue #267, opened 2024-08-28
Consolidate DatabaseProxy abstraction layer


dmohns:
**Background:**

In the multi-tenancy layout of the MPM there is some redundancy of User accounts and DatabaseProxy. User accounts "live" in the tenant databases. However, they need to get elevated and duplicated into DatabaseProxy. As a result the `DatabaseProxy` is essentially a UNION of all tenant user tables.

This is redundant and creates confusion.

**Scope:**

Please make User objects live at the MPM root. And use the DatabaseProxy work with the newly created User objects. Adapt services accordingly.

**Definition of done:**

- The model `DatabaseProxy` and all related services and code is removed
- The model `User` now lives at root level

dmohns:
Hey @beesaferoot 

I think this was closed by mistake, right 😅 ? Can we re-open it?

beesaferoot:
Oh right, I changed the status to closed my project list. Sorry

PR #461, opened 2025-01-18
refactor: Consolidate dbproxy and user models

beesaferoot:

With this refactor we migrate User model to the root micropowermanager database. Following, we consolidate database proxy model by resolving company id from user table directly from database proxy services.

Consideration for this refactor is that, admin users can see every user on the users table from the user management admin page.

This might have the effect or leaking admin accounts to other admin user.

Closes: #267 

### Pull Request checklist


munyanezaarmel (dismissed):

beesaferoot:
@dmohns  Can you take a look at this when you find the time. Thanks

beesaferoot:
> Consideration for this refactor is that, admin users can see every user on the users table from the user management admin page.
This might have the effect or leaking admin accounts to other admin user.


Update on this: this has been resolved by filtering the users based on their associated company id, that is admin user can only see users associated with the company id.  

@dmohns

src/backend/database/migrations/2025_01_18_100625_create_user_table.php
```diff
@@ -0,0 +1,33 @@
...
+            $table->increments('id');
+            $table->string('name');
+            $table->integer('company_id')->unsigned();
+            $table->string('email')->unique();
+            $table->string('password');
+            $table->rememberToken();
+            $table->timestamps();
+        });
+    }
+
```
dmohns, line 24: If our migrations "just" create the table this will break an existing installation of MPM, right? I.e. the new table will be empty.

Basically, here we would need a step that makes sure the new table is a UNION of all existing tables. Something like `INSERT INTO users ...`

src/backend/app/Services/DatabaseProxyService.php
```diff
@@ -2,27 +2,28 @@
 
 namespace App\Services;
 
-use App\Models\DatabaseProxy;
+use App\Models\User;
 use App\Services\Interfaces\IBaseService;
 use Illuminate\Database\Eloquent\Collection;
```
dmohns, line 7: I'm surprised to see this still being here. I was expecting `DatabaseProxyService` to be removed as well

src/backend/app/modules/DatabaseProxy/DatabaseProxyManagerService.php
```diff
@@ -5,20 +5,20 @@
...
         private CompanyDatabase $companyDatabase,
     ) {}
 
-    public function findByEmail(string $email): DatabaseProxy {
-        return $this->databaseProxy->findByEmail($email);
+    public function findByEmail(string $email): User {
+        return $this->user->findByEmail($email);
     }
 
     public function runForCompany(int $companyId, callable $callable) {
```
dmohns, line 24: I think we need `runForCompany` and `buildDatabaseConnection` from this model. I think these should go to the `CompanyService`. WDYT?

dmohns:
One more general comment: I see a couple of code sections being commented out. Is this still W.I.P? If not, please just remove them entirely 🔥

src/backend/app/Services/DatabaseProxyService.php
beesaferoot: I wasn't sure the best place to move some needed logic that lived inside `DatabaseProxyService`. 

From your comment below, I will move them to `CompanyService` and remove this entirely.

src/backend/app/modules/DatabaseProxy/DatabaseProxyManagerService.php
beesaferoot: That's a good place for it. Thanks for pointing that out!

src/backend/database/migrations/2025_01_18_100625_create_user_table.php
beesaferoot: > If our migrations "just" create the table this will break an existing installation of MPM, right?

I believe the demo company db will be dropped and reseeded to get things work. Would that be an issue?

beesaferoot: This totally makes sense. I have updated to accord for this. Please have a look

src/backend/database/migrations/tenant/2025_01_18_100835_drop_user_table.php
```diff
@@ -0,0 +1,55 @@
...
+                    DB::connection('micro_power_manager')->table('users')->insert([
+                        'id' => $user->id,
+                        'name' => $user->name,
+                        'company_id' => $user->company_id,
+                        'email' => $user->email,
+                        'password' => $user->password,
+                        'created_at' => $user->created_at,
+                        'updated_at' => $user->updated_at,
+                    ]);
+                }
```
dmohns, line 32: Could you add an `else` block here that makes this section fail if the core schema doesn't have the `users` table? Basically, we want to prevent the `tenant.users` table to be dropped if `micro_power_manager.users` doesn't exist.
This could happen, if for some reason the Tenant migration is run, before the root migration is run.

src/backend/database/migrations/tenant/2025_01_18_100835_drop_user_table.php
```diff
@@ -0,0 +1,55 @@
...
+
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up() {
+        // Check if the users table exists in the tenant connection
+        if (Schema::connection('tenant')->hasTable('users')) {
```
dmohns, line 16: Nit: I don't think we need this. If `tenant.users` doesn't exist at the time the migration is run, then something is fundamentally wrong and it's ok if application of the migration fails.

src/backend/app/Models/User.php
```diff
@@ -31,7 +31,7 @@ class User extends Authenticatable implements JWTSubject {
     public const COL_COMPANY_ID = 'company_id';
 
     public function __construct(array $attributes = []) {
-        $this->setConnection('tenant');
+        // $this->setConnection('tenant');
```
dmohns, line 34: Let's remove this entirely.

dmohns (commented):
Looks good nice. Two more small comments

beesaferoot:
Updated, kindly review @dmohns

src/backend/database/migrations/tenant/2025_01_18_100835_drop_user_table.php
```diff
@@ -0,0 +1,55 @@
...
+                    'name' => $user->name,
+                    'company_id' => $user->company_id,
+                    'email' => $user->email,
+                    'password' => $user->password,
+                    'created_at' => $user->created_at,
+                    'updated_at' => $user->updated_at,
+                ]);
+            }
+        } else {
+            throw new Exception('micro_power_manager.users table does not exist');
```
dmohns, line 32:
```suggestion
            throw new Exception('Table `micro_power_manager.users` does not exist. Please make sure the Central Database migration have been run.');
```

dmohns (approved):
Let's gooo :shipit:

dmohns (changes requested):
There is one more thing I want to check.

dmohns:
I think we also need to deal with the Relations of the `User` object.

Currently, the Tenant User object has the following relations

```
  Relations ..............................................................................................................................  
  address MorphOne ............................................................................................ App\Models\Address\Address  
  balanceHistory HasMany .................................................................................. App\Models\AgentBalanceHistory  
  assignedAppliance HasMany ........................................................................... App\Models\AgentAssignedAppliances  
  relationTicketUser HasOne ............................................................................. Inensus\Ticket\Models\TicketUser  
  notifications MorphMany .................................................................. Illuminate\Notifications\DatabaseNotification 
```

And we have to identify what to do with them.

beesaferoot:
> I think we also need to deal with the Relations of the `User` object.
> 
> Currently, the Tenant User object has the following relations
> 
> ```
>   Relations ..............................................................................................................................  
>   address MorphOne ............................................................................................ App\Models\Address\Address  
>   balanceHistory HasMany .................................................................................. App\Models\AgentBalanceHistory  
>   assignedAppliance HasMany ........................................................................... App\Models\AgentAssignedAppliances  
>   relationTicketUser HasOne ............................................................................. Inensus\Ticket\Models\TicketUser  
>   notifications MorphMany .................................................................. Illuminate\Notifications\DatabaseNotification 
> ```
> 
> And we have to identify what to do with them.

Haven't thought this far ahead 🤔; do you anticipate any issues interacting with the User object as is?

dmohns:
> Haven't thought this far ahead 🤔; do you anticipate any issues interacting with the User object as is?

To be honest, I don't know how this will work if we are building relations between Central and Tenant database models. However, we might not be required to it as often:

- I think all the Agent related stuff can simply be removed. `address`, `balanceHistory` and `assignedAppliance ` can simply go. They are already on the Agent model.
- The `TickerUser` should be fine. Because `ticket_user` has `user_id` column. The only problem might be, that the `user_id` might change when users get moved from Tenant to Central
- The `notification` relation I don't really understand. I guess we need to move this relation to Central as well. But I'm not sure if we are even using it.

beesaferoot:
> > Haven't thought this far ahead 🤔; do you anticipate any issues interacting with the User object as is?
> 
> To be honest, I don't know how this will work if we are building relations between Central and Tenant database models. However, we might not be required to it as often:
> 
> * I think all the Agent related stuff can simply be removed. `address`, `balanceHistory` and `assignedAppliance ` can simply go. They are already on the Agent model.
> * The `TickerUser` should be fine. Because `ticket_user` has `user_id` column. The only problem might be, that the `user_id` might change when users get moved from Tenant to Central
> * The `notification` relation I don't really understand. I guess we need to move this relation to Central as well. But I'm not sure if we are even using it.


`UserAddressController` uses the address relation to the user, should that also go?

dmohns:
> > > Haven't thought this far ahead 🤔; do you anticipate any issues interacting with the User object as is?
> > 
> > 
> > To be honest, I don't know how this will work if we are building relations between Central and Tenant database models. However, we might not be required to it as often:
> > 
> > * I think all the Agent related stuff can simply be removed. `address`, `balanceHistory` and `assignedAppliance ` can simply go. They are already on the Agent model.
> > * The `TickerUser` should be fine. Because `ticket_user` has `user_id` column. The only problem might be, that the `user_id` might change when users get moved from Tenant to Central
> > * The `notification` relation I don't really understand. I guess we need to move this relation to Central as well. But I'm not sure if we are even using it.
> 
> `UserAddressController` uses the address relation to the user, should that also go?
> 
> It doesn't seem to be used on the frontend though; is there a need to get a user's address?

Good find. I think it can `UserAddressController` can go yeah 🧹

dmohns (approved):
Let's see what happens 🤞 🙈 :shipit:

Issue #432, opened 2024-12-19
[Bug]: Cluster Dashboard timerange filtering in Financial Overview isn't working as expected


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Expected Behavior

The display of Financial data in the Dashboard should work intuitive.

### Actual Behavior

There are multiple behaviours which make interaction very unintuitive.

**Loading of initial data requires click of a button:**
When first opening the Dashboard no revenue data is displayed. Users first have to click on "Refresh Data". Note, this behaviour doesn't always apply. It's unclear when it happenes.

**Initial time range does not correspond to displayed data:**
<img width="1274" alt="Image" src="https://github.com/user-attachments/assets/b720fe21-238e-461b-b087-4b74de6d61e4" />

The initial time range is set to 3 month, but the line chart displays 12 month.

**Filtering only applies to linechart:**

When the time frame is changed (for example from 3 to 6 month) only the line chart gets adapted. The other charts stay the same. As a user I would expect the filter to apply to all three charts.

### Steps to reproduce

See above

### What operating system(s) are you using?

macOS

PR #471, opened 2025-02-03
fix Cluster Dashboard timerange filtering in Financial Overview 

munyanezaarmel:
… working as expected

This PR addresses the unintuitive behavior of the financial data display in the Dashboard.  It resolves three key issues:

1. **Initial Data Loading:** The "Refresh Data" button is no longer required for initial data loading.  The Dashboard now automatically fetches and displays the most recent data upon opening.  This provides a more seamless user experience and eliminates the confusion surrounding the button's inconsistent behavior.  The underlying logic for data fetching has been reviewed and optimized to ensure consistent and reliable initial loading.

2. **Consistent Time Range:** The initial time range for all charts is now consistently set to 3 months, matching the default selection.  This eliminates the discrepancy between the selected range and the displayed data in the line chart.  The time range selection now correctly drives the data displayed in all three charts.

3. **Unified Filtering:**  The time range filter now applies consistently to all three charts.  Changing the time frame now updates the data displayed in all charts, providing a unified and expected filtering experience.

closes: #432 

### Pull Request checklist


beesaferoot:
A quick on the dashboard shows some data isn't displayed (top row summary). Can you check why?


Screenshot 2025-02-03 at 10 33 28 AM

dmohns:
> A quick check on the dashboard shows some data isn't displayed (top row summary). Can you check why?
> 
> Screenshot 2025-02-03 at 10 33 28 AM

Yeah, same issue for me! On initial load the top row is empty.

I also noticed, that on initial load the map is not populated. Only after a couple of refreshes/button clicks this happens.

beesaferoot:
I figured the issue was from the backend; the endpoint `/api/dashboard/clusters` GET request only uses the cache. So the first request usually will not have the cached data, which causes the server to respond with empty data.


I traced it to the `ClustersDashboardCacheDataController`


Modifying the logic to ensure the cache has been populated before returning resolves the issue. 

```php

 public function index() {
        $cachedData = $this->clustersDashboardCacheDataService->getData();
        if (empty($cachedData)) {
            $this->clustersDashboardCacheDataService->setData();
        }

        return ApiResource::make($this->clustersDashboardCacheDataService->getData());
    }
```

@munyanezaarmel @dmohns

munyanezaarmel:
Thanks, @beesaferoot! I was thinking the same thing that the issue might be coming from the backend. Initially, I thought the first request was taking too long, causing a timeout and resulting in a delay, which is why you had to send a second request (by clicking the "Reload Data" button). Thank you so much for looking into this! You can check it again now.

cc: @dmohns

munyanezaarmel:
@dmohns, what do you mean by "map is not populated"? Could you provide more context?

dmohns:
> fix clusterdashboard cached data

Great find! Out of curiosity... Where you able to understand "why" we are doing that caching in the first place? It seems like the cluster data wouldn't be that massive 🤔

dmohns:
> @dmohns, what do you mean by "map is not populated"? Could you provide more context?

image

This is what it looks like to me ☝️ The map is empty and doesn't show any clusters/minigrids

munyanezaarmel:
> > @dmohns, what do you mean by "map is not populated"? Could you provide more context?
> 
> image
> This is what it looks like to me ☝️ The map is empty and doesn't show any clusters/minigrids

@dmohns, this is how it appears on my end. Could you confirm that you're on the correct branch? I also noticed that the financial overview data has not yet been fixed on your side
image

dmohns (approved):
This is an amazing change, thanks both of you 🙏

munyanezaarmel:
@dmohns, here is the issue. You can check it and let me know your thoughts: #483

dmohns:
> @dmohns, here is the issue. You can check it and let me know your thoughts: #483

Nice, thanks you. Let's merge this one then, it's approved.

PR #486, opened 2025-02-18
Update Documentation

93Kamuran:
### Pull Request checklist


dmohns:
Hey @93Kamuran 

I see you are essentially deleting all the documentation for Environment variables. But I don't understand why 🙈 

It looks to me like these **are** used. And, not only that they are crucial to configuring MPM.

Let's use `MAIL_` as an example... As a administrator of an MPM instance, how should I configure the email sending for my instance when not using the corresponding environment variables?

Instead of deleting these could you please instead update the documentation to actually reflect what these environment variables do?

For example `SUNKING_API_URL`... This ENV is the default URL that get's used when the SUNKING plugin get's enabled. It doesn't look like the user can customise this in the plugin view.

93Kamuran:
> Hey @93Kamuran
> 
> I see you are essentially deleting all the documentation for Environment variables. But I don't understand why 🙈
> 
> It looks to me like these **are** used. And, not only that they are crucial to configuring MPM.
> 
> Let's use `MAIL_` as an example... As a administrator of an MPM instance, how should I configure the email sending for my instance when not using the corresponding environment variables?
> 
> Instead of deleting these could you please instead update the documentation to actually reflect what these environment variables do?
> 
> For example `SUNKING_API_URL`... This ENV is the default URL that get's used when the SUNKING plugin get's enabled. It doesn't look like the user can customise this in the plugin view.

We have "mail_settings" table in tenant databases. The missing part is; We need to add one more settings tab for email settings on settings page. (We can not send all emails in the system for all tenants from one email address right?)

If you check the other areas I removed essipecially related plugins like "api url", we have default value on plugin table migrations.

dmohns:
> If you check the other areas I removed essipecially related plugins like "api url", we have default value on plugin table migrations.

Yeah, but the default value is taken from `ENV`. For example SUNKING: https://github.com/EnAccess/micropowermanager/blob/main/src/backend/database/migrations/micropowermanager/2022_14_12_create_sun_king_tables.php#L13

If the ENV is empty, the table will also be empty. Or am I missing something?

dmohns:
> We have "mail_settings" table in tenant databases. The missing part is; We need to add one more settings tab for email settings on settings page. (We can not send all emails in the system for all tenants from one email address right?)

Mail is a bit more complicated I think. There should be two mail settings.

First one is for instance level email to clients/tenants. I.e. if users sign up so they get Welcome email. This looks like it's actually hardcoded in the code: https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/Helpers/MailHelper.php#L19

Second one, is client emails to customers. I'm not even so sure this is a relevant use-case? I was thinking the client<>customer communication is done via SMS exclusively?

93Kamuran:
> > We have "mail_settings" table in tenant databases. The missing part is; We need to add one more settings tab for email settings on settings page. (We can not send all emails in the system for all tenants from one email address right?)
> 
> Mail is a bit more complicated I think. There should be two mail settings.
> 
> First one is for instance level email to clients/tenants. I.e. if users sign up so they get Welcome email. This looks like it's actually hardcoded in the code: https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/Helpers/MailHelper.php#L19
> 
> Second one, is client emails to customers. I'm not even so sure this is a relevant use-case? I was thinking the client<>customer communication is done via SMS exclusively?

updated

93Kamuran:
> > If you check the other areas I removed essipecially related plugins like "api url", we have default value on plugin table migrations.
> 
> Yeah, but the default value is taken from `ENV`. For example SUNKING: https://github.com/EnAccess/micropowermanager/blob/main/src/backend/database/migrations/micropowermanager/2022_14_12_create_sun_king_tables.php#L13
> 
> If the ENV is empty, the table will also be empty. Or am I missing something?

solved that with an update.

dmohns:
> solved that with an update.

Unfortunately, it doesn't work like this 🙈  

Few things:

- Changing "past" migrations is extremely dangerous it can lead to highly inconsistent situations. We should only do this in exceptional cases.
- In this specific case: When you remove the default value being read from ENV there is no way to use the production URL anymore. If I deploy the prod version, how would I customise it to point to SunKing live env?

Also, the scope of this PR is documentation only. In here, let's try to only document the status-quo. If you want to improve the way these are handled, please send a separate PR for that.

93Kamuran:
> > solved that with an update.
> 
> Unfortunately, it doesn't work like this 🙈
> 
> Few things:
> 
> * Changing "past" migrations is extremely dangerous it can lead to highly inconsistent situations. We should only do this in exceptional cases.
> * In this specific case: When you remove the default value being read from ENV there is no way to use the production URL anymore. If I deploy the prod version, how would I customise it to point to SunKing live env?
> 
> Also, the scope of this PR is documentation only. In here, let's try to only document the status-quo. If you want to improve the way these are handled, please send a separate PR for that.

Changing "past" migrations is extremely dangerous it can lead to highly inconsistent situations. We should only do this in exceptional cases.
   
     - Thank you for telling this to me :) If I'm not wrong there is no deployed version of the project still. Anyway  we can create an update migration or revert SunKing url in environment_variables.md file. Just tell me your preference please.
     
Also, the scope of this PR is documentation only. In here, let's try to only document the status-quo. If you want to improve the way these are handled, please send a separate PR for that. 
   
    -  So if you choose first option please create a task for this and let me create an update migration.

93Kamuran:
> > We have "mail_settings" table in tenant databases. The missing part is; We need to add one more settings tab for email settings on settings page. (We can not send all emails in the system for all tenants from one email address right?)
> 
> Mail is a bit more complicated I think. There should be two mail settings.
> 
> First one is for instance level email to clients/tenants. I.e. if users sign up so they get Welcome email. This looks like it's actually hardcoded in the code: https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/Helpers/MailHelper.php#L19
> 
> Second one, is client emails to customers. I'm not even so sure this is a relevant use-case? I was thinking the client<>customer communication is done via SMS exclusively?

DESCRIBE mail_settings;

  id	
  mail_host	
  mail_port	
  mail_encryption		
  mail_username		
  mail_password		
  created_at		
  updated_at		
  
Each tenant should specify its on mail settings, for an email which is gonna sent by MPM behalf tenant. MPM needs a email settings page/section  also email sending functions should be updated accordingly.

dmohns:
> > > We have "mail_settings" table in tenant databases. The missing part is; We need to add one more settings tab for email settings on settings page. (We can not send all emails in the system for all tenants from one email address right?)
> > 
> > 
> > Mail is a bit more complicated I think. There should be two mail settings.
> > First one is for instance level email to clients/tenants. I.e. if users sign up so they get Welcome email. This looks like it's actually hardcoded in the code: https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/Helpers/MailHelper.php#L19
> > Second one, is client emails to customers. I'm not even so sure this is a relevant use-case? I was thinking the client<>customer communication is done via SMS exclusively?
> 
> DESCRIBE mail_settings;
> 
> id mail_host mail_port mail_encryption mail_username mail_password created_at updated_at
> 
> Each tenant should specify its on mail settings, for an email which is gonna sent by MPM behalf tenant. MPM needs a email settings page/section also email sending functions should be updated accordingly.

My question is: Form the code it looks like the `mail_settings` is not used, but the `config` object is (and config object will be controlled via env variables. Can you point me to the place where the values from the table are used?

93Kamuran:
> My question is: Form the code it looks like the `mail_settings` is not used, but the `config` object is (and config object will be controlled via env variables. Can you point me to the place where the values from the table are used?

Yes, you're right! I re-checked the source code and noticed that we use mailing only for sending emails to tenant admins. and that functions use configs from config/mail.php. 

"mail_settings" may stay for tenant databases. We may consider using it in mailing operations between tenant & end customers in future. 

So I'm gonna revert my changes related mail envs.

docs/installation/environment-variables.md
```diff
@@ -142,87 +142,10 @@ Configure the following environment variable to enable MicroPowerManager to send
...
-#### Airtel
-
-For detailed information see [Airtel Developer Documentation](https://developers.airtel.africa/developer).
-
-| Environment Variable | Default                            | Description                                 |
-| -------------------- | ---------------------------------- | ------------------------------------------- |
-| `AIRTEL_REQUEST_URL` | **Required** (when plugin is used) | The Airtel service URL.                     |
-| `AIRTEL_USERNAME`    | **Required** (when plugin is used) | The Airtel username.                        |
-| `AIRTEL_PASSWORD`    | **Required** (when plugin is used) | The Airtel password.                        |
-| `AIRTEL_IPS`         | `[]`                               | List of IP whitelisted for Airtel services. |
```
dmohns, line 154: These seem used.

- https://github.com/EnAccess/micropowermanager/blob/main/src/backend/config/services.php#L35-L40
- https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/modules/Transaction/Provider/AirtelTransactionProvider.php#L46-L56

please don't remove them

docs/installation/environment-variables.md
```diff
@@ -142,87 +142,10 @@ Configure the following environment variable to enable MicroPowerManager to send
...
-
-#### Calin STS Meters
-
-For detailed information see [Calin Meter Developer Documentation](https://www.szcalinmeter.com/).
-
-| Environment Variable | Default                            | Description                                              |
-| -------------------- | ---------------------------------- | -------------------------------------------------------- |
-| `CALIN_CLIENT_URL`   | **Required** (when plugin is used) | Calin Meter client URL used for generating STS tokens.   |
-| `CALIN_USER_ID`      | **Required** (when plugin is used) | Calin Meter API username used for generating STS tokens. |
-| `CALIN_KEY`          | **Required** (when plugin is used) | Calin Meter API key used for generating STS tokens.      |
```
dmohns, line 164: These seem used

- https://github.com/EnAccess/micropowermanager/blob/main/src/backend/config/services.php#L43-L53
- https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/ManufacturerApi/CalinReadMeter.php#L29-L35

Please don't remove them

docs/installation/environment-variables.md
```diff
@@ -142,87 +142,10 @@ Configure the following environment variable to enable MicroPowerManager to send
...
-
-| Environment Variable           | Default                            | Description                                                               |
-| ------------------------------ | ---------------------------------- | ------------------------------------------------------------------------- |
-| `CALIN_SMART_COMPANY_NAME`     | **Required** (when plugin is used) | Calin Smart Meter company name used for communication with Calin API.     |
-| `CALIN_SMART_PURCHASE_API_URL` | **Required** (when plugin is used) | Calin Smart Meter Purchase API URL used for communication with Calin API. |
-| `CALIN_SMART_CLEAR_API_URL`    | **Required** (when plugin is used) | Calin Smart Meter Clear API URL used for communication with Calin API.    |
-| `CALIN_SMART_USER_NAME`        | **Required** (when plugin is used) | Calin Smart Meter username used for communication with Calin API.         |
-| `CALIN_SMART_PASSWORD`         | **Required** (when plugin is used) | Calin Smart Meter password used for communication with Calin API.         |
-| `CALIN_SMART_PASSWORD_VENT`    | **Required** (when plugin is used) | Calin Smart Meter password vent used for communication with Calin API.    |
```
dmohns, line 186: Please also remove the corresponding section in `services.php`. https://github.com/EnAccess/micropowermanager/blob/main/src/backend/config/services.php#L55-L64

docs/installation/environment-variables.md
```diff
@@ -142,87 +142,10 @@ Configure the following environment variable to enable MicroPowerManager to send
...
-| Environment Variable            | Default                            | Description                 |
-| ------------------------------- | ---------------------------------- | --------------------------- |
-| `VODACOM_SPID`                  | **Required** (when plugin is used) | Vodacom SPID.               |
-| `VODACOM_SPPASSWORD`            | **Required** (when plugin is used) | Vodacom SPPASSWORD.         |
-| `VODACOM_IPS`                   | **Required** (when plugin is used) | Vodacom IPs.                |
-| `VODACOM_REQUEST_URL`           | **Required** (when plugin is used) | Vodacom Request API URL.    |
-| `VODACOM_BROKER_CRT`            | **Required** (when plugin is used) | Path to broker `.crt` file. |
-| `VODACOM_SLL_KEY`               | **Required** (when plugin is used) | Path to `.key` file.        |
-| `VODACOM_CERTIFICATE_AUTHORITY` | **Required** (when plugin is used) | Path to `.cer` file.        |
-| `VODACOM_SSL_CERT`              | **Required** (when plugin is used) | Path to `.pem` file.        |
```
dmohns, line 202: These seem used

- https://github.com/EnAccess/micropowermanager/blob/main/src/backend/config/services.php#L22-L33
- https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/modules/Transaction/Provider/VodacomTransactionProvider.php#L311-L323

Please don't remove them

docs/installation/environment-variables.md
```diff
@@ -142,87 +142,10 @@ Configure the following environment variable to enable MicroPowerManager to send
...
-
-#### SunKing
-
-For detailed information see [SunKing Developer Documentation](https://sunking.com/)
-
-| Environment Variable | Default                            | Description                                                                      |
-| -------------------- | ---------------------------------- | -------------------------------------------------------------------------------- |
-| `SUNKING_API_URL`    | **Required** (when plugin is used) | SunKing API URL. For example `https://dev.assetcontrol.central.glp apps.com/v2`. |
-
```
dmohns, line 211: This has to stay for now, see below

docs/installation/environment-variables.md
93Kamuran: AirtelTransactionProvider should be removed. We need to develop a plugin for Airtel If we're planning that in future.

docs/installation/environment-variables.md
93Kamuran: CalinReadMeter.php is not using. We have plugins for Calin & Calin Smart meters. The code sample you shared me is legacy code from very old versions of MPM.  So that can be removed.

docs/installation/environment-variables.md
93Kamuran: CalinReadMeter.php is not using. We have plugins for Calin & Calin Smart meters. The code sample you shared me is legacy code from very old versions of MPM. So that can be removed.

docs/installation/environment-variables.md
93Kamuran: VodacomTransactionProvider should be removed. We need to develop a plugin for Airtel If we're planning that in future.

docs/installation/environment-variables.md
93Kamuran: ok

src/backend/database/migrations/micropowermanager/2022_14_12_create_sun_king_tables.php
```diff
@@ -10,7 +10,7 @@ public function up() {
             Schema::connection('tenant')->create('sun_king_api_credentials', static function (Blueprint $table) {
                 $table->increments('id');
                 $table->string('auth_url')->default('https://auth.central.glpapps.com/auth/realms/glp-dev/protocol/openid-connect/token');
-                $table->string('api_url')->default(config('services.sunKing.url'));
+                $table->string('api_url')->default('https://assetcontrol.central.glpapps.com/v2');
```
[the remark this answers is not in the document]
93Kamuran: This is fixed value, So we need to set it into "sun_king_api_credentials" table with default value once "install plugin" command runs. We do not need to show this informayion on UI side.

dmohns:
Hey @93Kamuran 

thanks for your responses. So, my point is: I don't want to remove these from the docs, as long as they are "used" (or mentioned, really) in the code. If I understand your comments correctly none of the code blocks where these appear is actually getting executed? Could you instead then simply remove the code (and the corresponding section of the docs)?

src/backend/database/migrations/micropowermanager/2022_14_12_create_sun_king_tables.php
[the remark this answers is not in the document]
dmohns: How would I (as a platform admin) then change my development environment to use the SunKing sandbox?

docs/installation/environment-variables.md
dmohns: Resolved here: https://github.com/EnAccess/micropowermanager/pull/526

docs/installation/environment-variables.md
```diff
@@ -142,87 +142,10 @@ Configure the following environment variable to enable MicroPowerManager to send
...
-| `CALIN_KEY`          | **Required** (when plugin is used) | Calin Meter API key used for generating STS tokens.      |
-
-If you meters are used which can send their consumption data to Calin's
-API the following environment variables need to be set.
-
-| Environment Variable | Default                            | Description                                                |
-| -------------------- | ---------------------------------- | ---------------------------------------------------------- |
-| `METER_DATA_URL`     | **Required** (when plugin is used) | Calin Meter API URL used for consumption data upload.      |
-| `METER_DATA_KEY`     | **Required** (when plugin is used) | Calin Meter API key used for consumption data upload.      |
-| `METER_DATA_USER`    | **Required** (when plugin is used) | Calin Meter API username used for consumption data upload. |
```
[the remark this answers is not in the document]
dmohns: They are still referenced in `.env.example` could you please remove them from there as well?

docs/installation/environment-variables.md
dmohns: Resolved here: https://github.com/EnAccess/micropowermanager/pull/526

dmohns (approved):

Issue #489, opened 2025-02-18
[Feature Request]: Vodacom Mozambique (M-Pesa) Generic C2B Integration


93Kamuran:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Implement the Vodacom Transaction API integration based on the provided specifications. The API consists of three main endpoints: validation, processing, and enquiry. All requests must be authenticated using a JWT token that includes expiresAt and companyId in the payload.

### Proposed Solution

Authentication
All requests must include a JWT token in the Authorization header as Bearer <token>.
The JWT payload should contain:
expiresAt
companyId

Endpoints

1. Validation

-     Request Type: POST
-     Endpoint: /api/vodacom/transactions/validation
-     Content-Type: application/json
-     Request Body:

```
{
  "serialNumber": "string", // Required
  "amount": 0.0, // Required (double)
  "payerPhoneNumber": "string", // Required
  "referenceId": "string" // Required (Unique Id)
}
```

Response:

-     Status: 200 OK
-     Response Body:

```
{
  "data": {},
  "success": true
}
```

2. Process Transaction

-     Request Type: POST
-     Endpoint: /api/vodacom/transactions/process
-     Content-Type: application/json

```
{
  "referenceId": "string", // Same as validation request (Required)
  "transactionId": "string" // ID of Vodacom transaction (Required)
}
```

Response:

-     Status: 200 OK
-     Response Body:

```
{
  "data": {
    "success": true
  }
}
```

3. Enquiry
(Vodacom should continue hitting this endpoint until the status is either Failed or Succeed.)
-     Request Type: POST
-     Endpoint: /api/vodacom/transactions/enquiry
-     Content-Type: application/json

```
{
  "referenceId": "string" // Same as validation request (Required)
}

```
Response:

-     Status: 200 OK
-     Response Body:

```
{
  "data": {
    "success": true,
    "status": ["Failed", "Succeed", "Pending"]
  }
}
```

Error Response:

-     Status: 400 Bad Request

```
{
  "data": {
[...]

PR #497, opened 2025-02-19
Create dummy vodacom endpoints

beesaferoot:

This PR adds dummy endpoints as described in #489 
### Pull Request checklist

### DIFF
--- a/src/backend/app/Http/Controllers/VodacomTransactionController.php
+++ b/src/backend/app/Http/Controllers/VodacomTransactionController.php
@@ -8,11 +8,7 @@ use App\Services\VodacomService;
 use Illuminate\Http\Request;
 
 class VodacomTransactionController extends Controller {
-    private VodacomService $vodacomService;
-
-    public function __construct(VodacomService $vodacomService) {
-        $this->vodacomService = $vodacomService;
-    }
+    public function __construct(private VodacomService $vodacomService) {}
 
     /**
      * Store a newly created resource in storage.
--- a/src/backend/app/Http/Controllers/VodacomTransactionController.php
+++ b/src/backend/app/Http/Controllers/VodacomTransactionController.php
@@ -43,16 +43,16 @@ class VodacomTransactionController extends Controller {
     }
 
     /**
-     * Validate Transaction
-     * 
-     * Validates a transaction before processing. Use this endpoint to verify if a transaction 
+     * Validate Transaction.
+     *
+     * Validates a transaction before processing. Use this endpoint to verify if a transaction
      * can proceed based on the provided information. This is typically the first step in the payment flow.
      *
      * @bodyParam serialNumber string required Unique identifier for the product/service being purchased pattern: ^[A-Z0-9]{8,12}$ Example: ABC123456789
      * @bodyParam amount number required Transaction amount in the local currency  Example: 15000
      * @bodyParam payerPhoneNumber string required Customer's phone number in international format pattern: ^258[0-9]{9}$ Example: 258712345678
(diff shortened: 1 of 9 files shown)


src/backend/app/Http/Controllers/VodacomTransactionController.php
```diff
@@ -2,10 +2,18 @@
...
 use App\Models\Transaction\VodacomTransaction;
+use App\Services\VodacomService;
 use Illuminate\Http\Request;
 
 class VodacomTransactionController extends Controller {
+    private VodacomService $vodacomService;
+
+    public function __construct(VodacomService $vodacomService) {
+        $this->vodacomService = $vodacomService;
+    }
```
dmohns, line 15: You can use (so-called) constructor property promotion here to simplify this to

```suggestion
    public function __construct(
        private VodacomService $vodacomService
    ) {}
```

beesaferoot: Updated.

dmohns:
One more general comment: For sharing this with Vodacom it would be important that this new endpoint shows up in our OpenAPI definitions. 

Could you ...

1. Check that `scribe` picks it up correctly? The process is semi-manual and documented here: https://github.com/EnAccess/micropowermanager/blob/main/docs/README.md#generate-openapi-docs
2. Add the updated OpenAPI to this PR as well?
3. Check if we have provided enough meta descriptions that the OpenAPI docs are meaningful to an external developer?

beesaferoot:
> One more general comment: For sharing this with Vodacom it would be important that this new endpoint shows up in our OpenAPI definitions.
> 
> Could you ...
> 
> 1. Check that `scribe` picks it up correctly? The process is semi-manual and documented here: https://github.com/EnAccess/micropowermanager/blob/main/docs/README.md#generate-openapi-docs
> 2. Add the updated OpenAPI to this PR as well?
> 3. Check if we have provided enough meta descriptions that the OpenAPI docs are meaningful to an external developer?

Will look into this.

beesaferoot:
> One more general comment: For sharing this with Vodacom it would be important that this new endpoint shows up in our OpenAPI definitions.
> 
> Could you ...
> 
> 1. Check that `scribe` picks it up correctly? The process is semi-manual and documented here: https://github.com/EnAccess/micropowermanager/blob/main/docs/README.md#generate-openapi-docs
> 2. Add the updated OpenAPI to this PR as well?
> 3. Check if we have provided enough meta descriptions that the OpenAPI docs are meaningful to an external developer?


Updated the OpenAPI spec, added endpoints descriptions. Please review  

@dmohns

beesaferoot:
> Nice, two comments on the OpenAPI
> 
> 1. Somehow the amount<>number data type is not correctly picked up. It says `any` and the example payload is also misleading. Could you look into that?
> 
> image
> 2. For the `payerPhoneNumber`. I would love to have a more precise example/documentation what is our understanding of "international format". Could you maybe look into how we can use a `pattern` here: https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings

The examples seems auto generated by scribe; there should be a way to get it to work correctly.

>  2. For the `payerPhoneNumber`. I would love to have a more precise example/documentation what is our understanding of "international format". Could you maybe look into how we can use a `pattern` here: https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings

I will do just. 


For the issue #497, is it possible we can include more details to the response and request? This will help make the documentation easier and perhaps better aligned.

dmohns (approved):

Issue #500, opened 2025-02-20
[Feature Request]: Move to MySQL rather than MariaDB


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

To have maximum flexibility we should support MySQL rather than MariaDB due more limitations in Cloud Hosting scenarios.

While the two are very similar they have some subtle differences which are actually breaking MPM when running with MySQL currently. Related to how JSON values are stored as `STRING` type in MariaDB and as `JSON` type in MySQL.

```sh
SQLSTATE[22032]: <<Unknown error>>: 3144 Cannot create a JSON value from a string with CHARACTER SET 'binary'.
at /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Connection.php:545)
```

### Proposed Solution

Drop Support for `MariaDB`

### Alternatives Considered

Continue supporting `MariaDB` and be less flexible with Cloud Hosting solutions.

PR #503, opened 2025-02-20
Make SQL statement MySQL complicated

dmohns:

Fixes

```
SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #1 of ORDER BY clause is not in GROUP BY clause and contains nonaggregated column 'DemoCompany_1.payment_histories.created_at' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```

when running with MySQL rather than MariaDB.

Makes progress on: #500 

### Pull Request checklist

### DIFF
--- a/src/backend/app/Models/PaymentHistory.php
+++ b/src/backend/app/Models/PaymentHistory.php
@@ -42,7 +42,7 @@ class PaymentHistory extends BaseModel {
             FROM payment_histories
             WHERE payer_id = :payer_id AND payer_type = :payer_type
             GROUP BY aperiod, payment_type
-            ORDER BY created_at {$order};
+            ORDER BY aperiod {$order};
             SQL;
 
         if ($limit !== null) {


dmohns:
cc @beesaferoot because you were having some trouble with this query a while ago 😅

beesaferoot:
> cc @beesaferoot because you were having some trouble with this query a while ago 😅

I forget why ordering by `created_at` fails. Does it have to do with a rule that enforces only fields that appear in the `GROUP BY` clause to be used in the `ORDER BY`?

dmohns:
> > cc @beesaferoot because you were having some trouble with this query a while ago 😅
> 
> I forget why ordering by `created_at` fails. Does it have to do with a rule that enforces only fields that appear in the `GROUP BY` clause to be used in the `ORDER BY`?

It didn't fail in the current setup. However, it does fail when trying to run the project with a MySQL database which (apparently) has much stricter rule set (which is good).

If you think about, this behaviour is ambiguous and should never have worked in the first place. `ORDER` happens after `GROUP` and we are referencing a non-grouped, non-aggregated column. Some more background here: https://dev.mysql.com/doc/refman/8.4/en/group-by-handling.html

PR #547, opened 2025-03-09
Fix larastan on level 2

### LINKED ISSUE #411: Fix `larastan` on level 2
**Scope:**

Currently `larastan` on level **2** reports errors which indicate broken code in our code base.

For each of the reported errors please decide, whether the corresponding code 

- should be removed as it's never used
- should be fixed as the application will break if executed

**Definition of done:**

The `larastan` level in `phpstan.neon` is set to **2** and the Github Action step `composer run larastan` passes.

alaincodes24:
This PR addresses the listed errors of `composer run larastan`

closes #411 


src/backend/app/Models/Meter/Meter.php
```diff
@@ -86,4 +87,8 @@ public function findBySerialNumber(string $meterSerialNumber): ?self {
     public function getId(): int {
         return $this->id;
     }
+
+    public function geo(): MorphOne {
+        return $this->morphOne(GeographicalInformation::class, 'owner');
```
dmohns, line 92: I don't think `Meter` is directly related to `GeoGraphicalInformation`. Instead it's `Address`<>`Device`<>`Meter`

src/backend/app/Models/Person/Person.php
```diff
@@ -125,4 +126,8 @@ public function livingInClusterQuery(int $clusterId) {
     public function getId(): int {
         return $this->id;
     }
+
+    public function meters() {
+        return $this->hasMany(Meter::class);
```
dmohns, line 131: The relation here is `Person`<>`Device`<>`Meter`

src/backend/app/Models/Transaction/Transaction.php
```diff
@@ -145,4 +153,8 @@ public function originalCash(): BelongsToMorph {
     public function originalWaveMoney(): BelongsToMorph {
         return BelongsToMorph::build($this, WaveMoneyTransaction::class, 'originalTransaction');
     }
+
+    public function meter() {
+        return $this->belongsTo(Meter::class, 'message', 'serial_number');
```
dmohns, line 158: The relation here is `Transaction`<>`Device`<>`Meter`

dmohns (commented):
Hey @alaincodes24 

Thank you very much, looks really good. I have left a few comments in the code.

I have a more general request (which is a little hard to explain, but let me try):
In this PR I would like us to not change any of the application logic. In particular, I see you are adding few relations here and there. I assume this is required because some pieces of the code are seemingly using these relations.

However, could we instead change the pieces of code that are using wrong references instead? I understand that some relations are waaaay to complicated (for example the whole `Address`<>`GeographicalInformation` block). If you notice some of these which could be improved, could you document your finding by creating an improvement ticket on Github?

So, in short, could you try to make it run on Larastan Level 2 without adding any relations?

src/backend/phpstan.neon
```diff
@@ -7,7 +7,7 @@ parameters:
         - app/
 
     # Level 9 is the highest level
-    level: 1
+    level: 2
```
dmohns, line 10: We can't do this yet, as not all level 2 errors will be resolved just yet. This will make our CI/CD fail.

dmohns (approved):
great work 👍

PR #568, opened 2025-03-18
Added Agent Sales Seeder + Fix Sold Appliance page

beesaferoot:

These changes allows us to generate data from agent to customer relationships, by which we can then query upon these interactions. 

The PR also includes minor big fixes on the UI related to appliance page.

### Pull Request checklist


dmohns (commented):
Hey @beesaferoot 

Thanks for the PR. I was not expecting this topic to be so big 🙈 

There is a couple of things that I don't like

- In the newly created `processSaleFromFactory` we are essentially "seeding" some data with things like `'tenure' => rand(6, 24),`. I know it's minor, but still. I'd love to have all seeding code in seeders.
- There is a lot of redundancy between `processSaleFromFactory` and `processSaleIfIsNotCreatedByFactory` which will lead to hard-to-maintain code.

Both of the above aren't really flaws of your solution. But I think there is a bigger underlying problem in the code base here:

 The `AgentSoldApplianceObserver` seems to heavily rely on `request()->input()`. This feels wrong on so many levels. 

- Currently the Observer implements a lot of business logic like creating model objects from requests and setting relation ships between them. By understanding is that Observer should _react_ to CRUD operations rather than _performing_ them.
- Observer cannot be used outside of `Request` context. This is exactly the problem you are facing, which leads to a lot of code duplication.

What do you think about the following suggestion:

1. Let's merge your changes the way they are, so we can benefit from the additional Seeding data generation.
2. Let's create a follow-up ticket to refactor `AgentSoldApplianceObserver` in a way that it doesn't require `request()->input().

Number 2. will be quite big, as it will require shuffling a lot of code around from the Observer into corresponding Services and Controllers. But I think it would be a valuable addition and improvement to the code base.

WDYT?

beesaferoot:
> Hey @beesaferoot
> 
> Thanks for the PR. I was not expecting this topic to be so big 🙈
> 
> There is a couple of things that I don't like
> 
> * In the newly created `processSaleFromFactory` we are essentially "seeding" some data with things like `'tenure' => rand(6, 24),`. I know it's minor, but still. I'd love to have all seeding code in seeders.
> * There is a lot of redundancy between `processSaleFromFactory` and `processSaleIfIsNotCreatedByFactory` which will lead to hard-to-maintain code.
> 
> Both of the above aren't really flaws of your solution. But I think there is a bigger underlying problem in the code base here:
> 
> The `AgentSoldApplianceObserver` seems to heavily rely on `request()->input()`. This feels wrong on so many levels.
> 
> * Currently the Observer implements a lot of business logic like creating model objects from requests and setting relation ships between them. By understanding is that Observer should _react_ to CRUD operations rather than _performing_ them.
> * Observer cannot be used outside of `Request` context. This is exactly the problem you are facing, which leads to a lot of code duplication.
> 
> What do you think about the following suggestion:
> 
> 1. Let's merge your changes the way they are, so we can benefit from the additional Seeding data generation.
> 2. Let's create a follow-up ticket to refactor `AgentSoldApplianceObserver` in a way that it doesn't require `request()->input().
> 
> Number 2. will be quite big, as it will require shuffling a lot of code around from the Observer into corresponding Services and Controllers. But I think it would be a valuable addition and improvement to the code base.
> 
> WDYT?

Thanks for the review @dmohns,  I agree it does need some refactoring and cleanup.


Although one of the reasons I opted to use an entirely different function for seeding purposes was because it wasn't clear why the `request` guard was important and I didn't want to go too deep while still focusing on the high-level task.

beesaferoot:
> I propose we move the seeding code inside seeders for better separation of concerns.

@alaincodes24  Do you mind clarifying a bit more on this?  Perhaps you can point to an example where the seed code is used outside a seeder.

beesaferoot:
Updated, kindly review @dmohns @alaincodes24

dmohns (commented):
This looks great. Two follow up questions:

1. When as we are changing the `request()->input()` logic I would want to be able to test that AgentSoldAppliance API endpoints still work as expected. Could you add an example HTTP collection with some dummy data, so we can test before and after the change?
2. I'm a bit confused as to why we need to add columns to `agent_sold_appliance` table. Where is this data stored right now?

beesaferoot:
> This looks great. Two follow up questions:
> 
> 1. When as we are changing the `request()->input()` logic I would want to be able to test that AgentSoldAppliance API endpoints still work as expected. Could you add an example HTTP collection with some dummy data, so we can test before and after the change?
> 2. I'm a bit confused as to why we need to add columns to `agent_sold_appliance` table. Where is this data stored right now?


I will be including a collection for this endpoint as noted. 

For 2) so from my investigation, fields like `down_payment` and `tenure` used in association with `asset` table. As you pointed out in your previous comment, the Observer logic needed these fields for handling model creation that needed this inform but were not available in the `AgentSoldAppliance` model, hence the need to reference the `request` directly. 

I considered two approaches:
1.) Provide these fields in the `agent_sold_appliances` table thereby removing a need for referencing request. 
2.) Provide these fields as temporary data in the `AgentSoldAppliance` model but not on the table. This require extra logic to achieve. 

After discusing a bit with Alain about a more idiomatic approach, I decided 1.) works well and is more laravel like.

beesaferoot:
Added sample endpoint (app) usage as bruno collections. 

@dmohns

dmohns:
I meant something slightly different 🙈 

We already have the columns `down_payment`, `tenure` (a.k.a `rate_count`) and `first_payment_date ` in our database. Namely, on the `asset_people` table.
image

It feels a bit wrong to add the same fields to another model (`agent_sold_appliances`). Because then we would need to make sure the two are in sync.

I understand why this is needed in the current approach: Because currently the `AgentSoldApplianceObserver` creates this `AgentPerson` object.

Essentially, what I'm wondering is if we should move the responsibilities of this current Observer into Controller and Services.

src/backend/database/seeders/AgentApplianceSalesSeeder.php
```diff
@@ -0,0 +1,71 @@
...
+        // Assign Assets to Agents
+        $assignedAppliances = $agents->flatMap(function ($agent) use ($assets, $customers) {
+            return AgentAssignedAppliances::factory()->count(3)->create([
+                'agent_id' => $agent->id,
+                'appliance_id' => $assets->random()->id,  // Use existing assets
+                'user_id' => $customers->random()->id,    // Use existing customers
+            ]);
+        });
+
+        // Simulate Sales and Trigger Observer
```
dmohns, line 53: This comment is no longer true, i.e. we no longer trigger the observer, just simulate the sales, right?

src/backend/app/Services/AgentSoldApplianceService.php
```diff
@@ -97,4 +109,97 @@ function ($q) use ($agentId) {
...
             ->paginate();
     }
+
+    public function getAgentsByCustomerId(int $customerId): Collection {
+        return Agent::whereHas('soldAppliances', function ($query) use ($customerId) {
+            $query->where('person_id', $customerId);
+        })->get();
+    }
+
+    public function processSale(AgentSoldAppliance $agentSoldAppliance, array $requestData = []): void {
```
dmohns, line 119: For some consistency with the rest of the code, could we call this function `processSaleFromRequest`?

dmohns (commented):
Looks good 👍 Two more small comments

src/backend/database/seeders/AgentApplianceSalesSeeder.php
beesaferoot: That's correct. Good catch, thanks!

src/backend/app/Services/AgentSoldApplianceService.php
beesaferoot: Updated.

dmohns (approved):
Really good work 🤩 Thanks for the patience!

PR #602, opened 2025-03-30
Fix larastan relationships errors on level 2 and 3

alaincodes24:
### Pull Request checklist


dmohns:
Hey @alaincodes24 

I'm having some trouble understanding this PR 🙈 It looks like most of the changes in here are getting reverted here: https://github.com/EnAccess/micropowermanager/pull/602 Can you explain briefly what the motivation for this PR is?

src/backend/app/modules/Apps/CustomerRegistration/CustomerRegistrationAppService.php
```diff
@@ -55,8 +55,9 @@ public function createCustomer(AndroidAppRequest $request): Person {
             'in_use' => 1,
         ];
         $meter = $this->meterService->create($meterData);
+        $personData = $person->toArray();
```
dmohns, line 58: I assume you do this to resolve

```
Access to an undefined property Illuminate\Database\Eloquent\Model::$id.
```

Whoever, here you are not really fixing the root cause of the issue, but the symptom. The underlying issue is in `PersonService.php`. 

```
public function createFromRequest(Request $request): Model
```

returns a generic `Model`. But should return `Person` instead.

src/backend/app/Services/MiniGridDashboardCacheDataService.php
```diff
@@ -152,8 +152,8 @@ public function setData($dateRange = []) {
             foreach ($period as $d) {
                 $day = $d->format('o-W');
                 foreach ($ticketCategories as $tC) {
-                    $result[$day][$tC->label_name]['opened'] = 0;
-                    $result[$day][$tC->label_name]['closed'] = 0;
+                    $result[$day][$tC['label_name']]['opened'] = 0;
+                    $result[$day][$tC['label_name']]['closed'] = 0;
```
dmohns, line 156: I don't think this works. `$tC->label_name` is the correct way to access the `label_name` property. However, Larastan is failing to understand that `label_name` is indeed a property of `ticketCategories`. 

What about 

```
/**
 * @property string $label_name
 */
class TicketCategory extends BaseModel {
    use HasFactory;

    protected $table = 'ticket_categories';
}
```

src/backend/app/Models/AssetPerson.php
```diff
@@ -47,4 +47,8 @@ public function creator(): MorphTo {
     public function device(): BelongsTo {
         return $this->belongsTo(Device::class);
     }
+
+    public function assetType(): BelongsTo {
+        return $this->belongsTo(AssetType::class);
+    }
```
dmohns, line 53: `AssetPerson` is not linked to `AssetType` directly. This tries to solve 

```
Access to an undefined property App\Models\AssetPerson::$assetType.
```

What about:

```
$loan->assetPerson->asset->assetType->name,
```

instead?

src/backend/app/Misc/TransactionDataContainer.php
```diff
@@ -43,11 +44,11 @@ public static function initialize(Transaction $transaction): TransactionDataCont
...
             $container->device = $deviceService->getBySerialNumber($transaction->message);
             $container->tariff = null;
-            $container->manufacturer = $container->device->device->manufacturer()->first();
+            $container->manufacturer = $container->device->device->manufacturer ?? null;
 
             if ($container->device->device_type === Meter::RELATION_NAME) {
                 $meter = $container->device->device;
-                $container->tariff = $meter->tariff()->first();
+                $container->tariff = $meter->tariff ?? null;
             }
```
dmohns, line 52: I think the issue here is, that Larastan doesn't understand that `$container->device->device` is a Meter in this `if` block.

What about we change this block to

```php
            if ($container->device->device instanceof Meter) {
                $meter = $container->device->device;
                $container->manufacturer = $meter->manufacturer()->first();
                $container->tariff = $meter->tariff()->first();
            }
```

instead?

dmohns: Also, probably want to add 

```
$container->meter = $meter;
```

Else the calls in `AssessRate.php` `$transactionData->meter` will probably fail.

src/backend/app/Http/Controllers/AuthController.php
```diff
@@ -82,7 +83,7 @@ protected function respondWithToken($token) {
             [
                 'access_token' => $token,
                 'token_type' => 'bearer',
-                'expires_in' => $guard->factory()->getTTL() * 60,
+                'expires_in' => JWTAuth::factory()->getTTL() * 60,
```
dmohns, line 86: This makes the block 5 lines above

```php
        /** @var JWTGuard $guard */
        $guard = auth()->guard('agent_api');
```

redundant?

src/backend/composer.lock
```diff
@@ -1831,45 +1831,6 @@
...
             "time": "2025-02-03T10:55:03+00:00"
         },
-        {
-            "name": "inensus/africas-talking",
-            "version": "dev-main",
-            "dist": {
-                "type": "path",
-                "url": "./packages/inensus/africas-talking",
-                "reference": "98c3cc2179e331b2ea3c657cf1b8631fe2feaa34"
-            },
```
dmohns, line 1841: I'm curious, why is this getting removed?

dmohns (commented):
Two more comments

src/backend/app/Http/Middleware/UserDefaultDatabaseConnectionMiddleware.php
```diff
@@ -89,14 +89,18 @@ private function handleApiRequest(Request $request, \Closure $next) {
...
+            $companyId = $guard->payload()->get('companyId');
             if (!is_numeric($companyId)) {
                 throw new \Exception('JWT is not provided');
             }
         } elseif ($this->resolveThirdPartyApi($request->path())) {
             $companyId = $this->apiCompanyResolverService->resolve($request);
         } else { // web client authenticated user requests
-            $companyId = auth('api')->payload()->get('companyId');
+            /** @var \Tymon\JWTAuth\JWTGuard */
+            $guard = auth('agent_api');
```
dmohns, line 102: Careful here ⚠️ From what I understand this is going to break the code.

We use two auth providers in this project https://github.com/EnAccess/micropowermanager/blob/main/src/backend/config/auth.php#L15-L22 

- `api` this uses the central `users` table for auth
- `agent_api` this uses the tenant's `agents` table for auth

Here you are essentially changing the auth provider. It's probably just a copy/paste error, but we need to be cautious.

dmohns (changes requested):

src/backend/app/Http/Middleware/UserDefaultDatabaseConnectionMiddleware.php
alaincodes24: @dmohns absolutely. thanks for pointing it out!

dmohns (approved):
Nice, thank you 🙏

Issue #608, opened 2025-04-01
[Feature Request]: Consistent display of currencies in Amount/Prices/Commissions etc...


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Currently in MicroPowerManager we use various ways to display currencies. 

For example:

- No currency displayed at all 

  <img width="624" alt="Image" src="https://github.com/user-attachments/assets/2138f8be-c013-468e-9dbf-155d32285f59" />

- Duplicate currency displayed as prefix and suffix

  <img width="622" alt="Image" src="https://github.com/user-attachments/assets/f88c6833-1870-4313-b047-b95efa453551" />

- Currency displayed as prefix with fractional digits always displayed

  <img width="627" alt="Image" src="https://github.com/user-attachments/assets/ea16585b-099c-4d2c-97c5-0ed7d303dfd7" />

- No currency displayed with fractional digits only displayed when non-zero

  <img width="635" alt="Image" src="https://github.com/user-attachments/assets/2d3427db-7a7d-41b8-b3ad-832eac70adb5" />

- Currency displayed as suffix with no space, but thousands separator
  
  <img width="647" alt="Image" src="https://github.com/user-attachments/assets/222dc621-d6fb-4f21-a530-a1fc1a957f46" />

This is bad user experiences. And specifically when no currency is displayed at all, can cause confusion.

### Proposed Solution

Please make sure all Amount/Prices/Commissions etc... are displayed in a consistent fashion. The formatting should be implemented in a single shared function such that it's easy to later change the formatting if we so choose.

Please start using the following style **Currency displayed as prefix with fractional digits only displayed when non-zero**.

### Alternatives Considered

N/A

### Additional Information

As this might be a big tasks please create sub-tickets as needed. This ticket will then track the overall progress.

PR #617, opened 2025-04-04
feat: Consistent display of currencies in Amount/Prices/Commissions

munyanezaarmel:
added Consistent display of currencies 

closes: #608
 the currency should displayed  as prefix with fractional digits
image

### Pull Request checklist

### DIFF
--- a/src/frontend/src/mixins/currency.js
+++ b/src/frontend/src/mixins/currency.js
@@ -1,28 +1,9 @@
 import store from "../store/store"
+import { readable } from "./numbers.js"
 
 export const currency = {
   methods: {
-    readable(amount, separator = ",") {
-      // Check for undefined or null amount and return 0
-      if (amount === undefined || amount === null) return "0"
-      // Convert the amount to a string
-      amount = amount.toString()
-      // If the amount is not a valid float, return it as is
-      if (
-        isNaN(parseFloat(amount)) ||
-        parseFloat(amount).toString() !== amount
-      ) {
-        return amount
-      }
-      // Split the amount into whole and decimal parts
-      let [whole, decimal] = amount.replace(/\s+/g, "").split(".")
-      // Format the whole number part with the separator
-      whole = whole.replace(/\B(?=(\d{3})+(?!\d))/g, separator)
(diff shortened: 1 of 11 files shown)


dmohns:
Hey @munyanezaarmel great stuff! I noticed a lot of code was using a function called `readable` before. As part of this PR could you please also remove this function, as it's no longer needed? https://github.com/EnAccess/micropowermanager/blob/main/src/frontend/src/mixins/currency.js#L5-L25

munyanezaarmel:
> Hey @munyanezaarmel great stuff! I noticed a lot of code was using a function called `readable` before. As part of this PR could you please also remove this function, as it's no longer needed? https://github.com/EnAccess/micropowermanager/blob/main/src/frontend/src/mixins/currency.js#L5-L25

Hey @dmohns , thanks for the feedback! Just wanted to double-check before removing the readable function — it's actually used for general number formatting (e.g. 10000 → 10,000), while moneyFormat is specifically for currency values with symbols and decimals.

They serve different use cases, so unless we're fully switching to moneyFormat everywhere or don't need plain number formatting anymore, readable might still be useful.

dmohns:
> > Hey @munyanezaarmel great stuff! I noticed a lot of code was using a function called `readable` before. As part of this PR could you please also remove this function, as it's no longer needed? https://github.com/EnAccess/micropowermanager/blob/main/src/frontend/src/mixins/currency.js#L5-L25
> 
> Hey @dmohns , thanks for the feedback! Just wanted to double-check before removing the readable function — it's actually used for general number formatting (e.g. 10000 → 10,000), while moneyFormat is specifically for currency values with symbols and decimals.
> 
> They serve different use cases, so unless we're fully switching to moneyFormat everywhere or don't need plain number formatting anymore, readable might still be useful.

Uff... yeah you are right. It's used in few places to format non-currency numbers. But this not ideal 😛 The function `readable` is located in a file called `currency.js`. Could you instead move the `readable` to it's own file, for example `numbers.js` and then adjust the code that uses to get it from there?

dmohns (approved):
Thanks!

Issue #616, opened 2025-04-03
[Feature Request]: Add API Endpoint for Fetching Solar Home System by ID


munyanezaarmel:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

We need an API endpoint to fetch details of a Solar Home System (SHS) by its ID. This is required to retrieve data for the newly added SHS Detail Page.

### Proposed Solution

**1. Backend**

- Implement a new GET /solarSystem/{id} endpoint.

The endpoint should return details such as:

- General Information (e.g., Manufacturer, Model, etc.)

- Customer Details (if applicable)

- Link to Sold Appliance (if the SHS is part of a deferred payment scheme).

**2. Integration**

- The UI for the SHS Detail Page has already been implemented.

- The frontend is waiting for this endpoint to display SHS details.

**3 .References**
PR: #615 (SHS Detail Page UI added, pending API integration).

**4 .Priority**
🚀 High (UI is blocked until this endpoint is available).

### Alternatives Considered

One alternative solution considered was retrieving solar system details through an existing generic endpoint that fetches multiple solar systems and filtering the data on the frontend. However, this approach would be inefficient, especially for large datasets, as it would require fetching unnecessary data and increase response time.

Another option was to cache frequently accessed solar system details on the frontend to reduce API calls. While this could improve performance, it wouldn't fully address the need for a dedicated endpoint that provides real-time, up-to-date information.

Ultimately, implementing a dedicated GET /solarSystem/{id} endpoint is the most efficient and scalable solution, ensuring direct and optimized access to the required data

dmohns:
Resolved by: https://github.com/EnAccess/micropowermanager/pull/619

PR #619, opened 2025-04-07
Add shs endpoint to get detailed info by id 

beesaferoot:

This PR implements endpoint to fetch details of a solar home system by it's id. 

Cloes: #616
### Pull Request checklist

### DIFF
--- a/src/backend/app/Http/Controllers/SolarHomeSystemController.php
+++ b/src/backend/app/Http/Controllers/SolarHomeSystemController.php
@@ -47,8 +47,6 @@ class SolarHomeSystemController extends Controller {
     }
 
     public function show($id): ApiResource {
-        $solarHomeSystem = $this->solarHomeSystemService->getById($id);
-
-        return ApiResource::make($solarHomeSystem->load(['manufacturer', 'appliance', 'device.person', 'device.address.geo']));
+        return ApiResource::make($this->solarHomeSystemService->getById($id));
     }
 }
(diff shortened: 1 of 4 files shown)


src/backend/app/Http/Controllers/SolarHomeSystemController.php
```diff
@@ -45,4 +45,10 @@ public function search(Request $request): ApiResource {
 
         return ApiResource::make($this->solarHomeSystemService->search($term, $paginate));
     }
+
+    public function show($id): ApiResource {
+        $solarHomeSystem = $this->solarHomeSystemService->getById($id);
+
+        return ApiResource::make($solarHomeSystem->load(['manufacturer', 'appliance', 'device.person', 'device.address.geo']));
```
dmohns, line 52: Is this really required here? I.e. doesn't all this information already get loaded in `solarHomeSystemService->getById($id);`? The `device.address.geo` isn't there, but can't we just add it to the service?

The background is: It would be nice if all the "logic" can be in the service. This would be consistent with most other "get by id" endpoints where they just call the corresponding service's `getById` method.

beesaferoot: Oh, rightly spotted. I will move it into the service

beesaferoot:
Updated

dmohns (approved):
Thanks 😉

PR #623, opened 2025-04-07
Fix sold appliances not clickable in agent detail view

### LINKED ISSUE #607: [Feature Request]: Sold Appliances not clickable in Agent Detail View
### Preflight Checklist

### Problem Description

As a admin I would like to explore the appliances sold by an Agent. Currently, I have to

- Go to the Agent Detail View and manually copy the Customer Name
- Go to the Customer Overview filter for the customer

### Proposed Solution

The Sold Appliance List in Agent Detail View should be clickable

Image

This should work exactly the same way as for exploring Sold Appliance from User Detail View. For example clicking on Sold Appliance id=5 should redirect to `/#/sold-appliance-detail/5`

Image

### Alternatives Considered

N/A

### Additional Information

_No response_

alaincodes24:
Fixed Sold Appliance not clickable in agent detail view

Closes #607 


beesaferoot (approved):

src/frontend/src/modules/Agent/Appliances/SoldApplianceList.vue
```diff
@@ -83,6 +85,9 @@ export default {
         this.agentSoldApplianceService.list.length,
       )
     },
+    detail(id) {
+      this.$router.push({ path: "/sold-appliance-detail/" + id })
+    },
```
dmohns, line 90: Hey @alaincodes24 

Rather than "hardcoding" the routes here, could you pull these routes in via the corresponding services/repositories? 

For example for SolarHomeSystems we would use https://github.com/EnAccess/micropowermanager/blob/main/src/frontend/src/repositories/SolarHomeSystemRepository.js#L10-L12 for SHS. This would be more consistent with how the rest of the frontend is structured.

alaincodes24: @dmohns  I checked the code for similar scenarios, it appears we follow this way in several other places. For instance, in `AgentList.vue` or `Clients.vue` we define a detail function which gets called on a click.
`
detail(id) {
      this.$router.push({ path: "/agents/" + id })
    },`
    
I thought I would clarify. Did you mean to move the path/route to the service and call it from the service or to move the whole `detail` function inside the service and call it from that very service? Thank you for clarifying.

dmohns: Ahh, my bad 🙈 I confused frontend and backend routes here. Backend routes (i.e. API paths) are managed in repository that I linked.
However, you are dealing with **frontend** routes. There is indeed no real convention how we manage these, rather than hardcoding the values as we need them. So, ignore my comment.

alaincodes24:
@dmohns I agree as well. I will go ahead and rename the method! I appreciate the feedback.

dmohns (approved):
Thanks a lot!

PR #660, opened 2025-04-15
Draft:

beesaferoot:
### Pull Request checklist


dmohns (commented):
More general question: Can't we just run the Test Suite with of `.env.micropowermanager-backend` ?

For example using https://github.com/c-py/action-dotenv-to-setenv

Then we don't need to deal with generating a key, and could also simplify the workflow?

beesaferoot:
> More general question: Can't we just run the Test Suite with of `.env.micropowermanager-backend` ?
> 
> For example using https://github.com/c-py/action-dotenv-to-setenv
> 
> Then we don't need to deal with generating a key, and could also simplify the workflow?

I will give this a try. I don't immediately see why we can't use this approach.

dmohns:
> > More general question: Can't we just run the Test Suite with of `.env.micropowermanager-backend` ?
> > For example using https://github.com/c-py/action-dotenv-to-setenv
> > Then we don't need to deal with generating a key, and could also simplify the workflow?
> 
> I will give this a try. I don't immediately see why we can't use this approach.

Only thing to have in mind is the `hostname` of the DB service. In our `.env` we set it to `db` but in the Github Action we set it to `mysel`. Probably need to adapt the service definition in the Action 👍

beesaferoot:
> > > More general question: Can't we just run the Test Suite with of `.env.micropowermanager-backend` ?
> > > For example using https://github.com/c-py/action-dotenv-to-setenv
> > > Then we don't need to deal with generating a key, and could also simplify the workflow?
> > 
> > 
> > I will give this a try. I don't immediately see why we can't use this approach.
> 
> Only thing to have in mind is the `hostname` of the DB service. In our `.env` we set it to `db` but in the Github Action we set it to `mysel`. Probably need to adapt the service definition in the Action 👍

So, I also think we might run into another issue. Laravel doesn't seem to consider the environment for key generation, it specifically wants to check for the `.env` file. 

This will help remove the duplicated environment values but not the key gen step.

PR #661, opened 2025-04-15
Add CI environment configuration and update test suite workflow

beesaferoot:
### Pull Request checklist


dmohns:
Following up the other thread here:

> So, I also think we might run into another issue. Laravel doesn't seem to consider the environment for key generation, it specifically wants to check for the .env file.

> This will help remove the duplicated environment values but not the key gen step.

I understand that `php artisan key:generate` requires an `.env` file. But here we are trying to achieve something slightly different. The reason for why we wanted to call `php artisan key:generate` in the first place, was because `APP_KEY` wasn't populated before. 
But with the new approach of exposing the entire `.env.micropowermanager-ci` file the `APP_KEY` environment variable should be present. And we no longer need to run `php artisan key:generate` to generate it.

Or am I missing something?

PR #663, opened 2025-04-16
Improve `Box` component

dmohns:

Couple of improvements to the `Box` component as a preparation for further adjustment and styling changes in the future.

- Remove duplicated code, which also causes CSS issues by affecting each other
- Remove wrong and unused color props
- Set color consistently via single prop

**Before:**

image

**After:**

image

### Pull Request checklist


munyanezaarmel:
@dmohns a screenshoot of the color changed might be helpful

dmohns:
Hey @munyanezaarmel 

This isn't changing any colours. For a user everything looks like before. The only difference is that we are now displaying the additional text. I will add a screen shot of that 👍

dmohns:
@munyanezaarmel added screenshot, if you can check

beesaferoot:
@dmohns Is the text colour change from the screenshots green to black intentional? i.e (0/0, TSZ)

dmohns:
> @dmohns Is the text colour change from the screenshots green to black intentional? i.e (0/0, TSZ)

Ohh, good catch, I forgot to mention:

Yeah, there is a slight change. As in the sub-text is now just simply a solid dark color for all boxes. Before this was highly inconsistent. 

Sometime we were using a dark variant of the box color. For example dark red here:

image

But I had to zoom in all the way, to even notice a difference.

Sometimes all greed

image


Sometimes a wild mix of colours

image

Additionally, there was some confusion because "subTextColor" was actually set by the "iconBoxColor" property...  And the "subTextColor" property was used, see https://github.com/EnAccess/micropowermanager/blob/main/src/frontend/src/modules/Box.vue#L20

So my reasoning here was: Let's make it simple and consistent for now. If we want to customise this later, we can do it in a more structured way than currently.

beesaferoot (approved):

Issue #665, opened 2025-04-16
[Bug]: Cannot run seeder with production images


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Expected Behavior

Running seeder should be possible even when using production images.

### Actual Behavior

```sh
[...]

PR #666, opened 2025-04-16
Fix seeders in production image

dmohns:

Closes: #665 

Essentially, we are moving `fakerphp/faker` to a non-dev dependency, such that we can use it in production images. This was introduced by https://github.com/EnAccess/micropowermanager/pull/625 where the install mode was changed to `--no-dev` in prod.

The use-case of running seeders for populate demo data seems to be a bit non-standard, but I think it makes sense in our setup.

### Pull Request checklist

### DIFF
--- a/docker/DockerfileBackendProd
+++ b/docker/DockerfileBackendProd
@@ -60,7 +60,7 @@ RUN chmod +x /usr/local/bin/entrypoint-backend-prod.sh
 
 # install MicroPowerManager
 WORKDIR /var/www/html
-RUN composer install --no-dev
+RUN composer install --no-dev --optimize-autoloader
 
 # define entrypoint
 ENTRYPOINT ["/usr/local/bin/entrypoint-backend-prod.sh"]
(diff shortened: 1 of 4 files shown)


beesaferoot:
@dmohns  Correct me, but my understanding was that we were already using seeding data for our demo version, no?
Maybe I am missing something. Is the production image different from what we run our demo version on?

dmohns:
> @dmohns Correct me, but my understanding was that we were already using seeding data for our demo version, no? Maybe I am missing something. Is the production image different from what we run our demo version on?

No, you are correct. Our Demo version is running production images (the one from DockerHub). And yes, we are using seeding in the Demo. But this was broken by: https://github.com/EnAccess/micropowermanager/pull/625

We never really identified, as in the Demo version we  don't "reset" the Demo as often. So, we haven't actually tried to re-seed Demo for a while. Trying to do this right now, yielded the error of the linked ticket.

beesaferoot:
> > @dmohns Correct me, but my understanding was that we were already using seeding data for our demo version, no? Maybe I am missing something. Is the production image different from what we run our demo version on?
> 
> No, you are correct. Our Demo version is running production images (the one from DockerHub). And yes, we are using seeding in the Demo. But this was broken by: #625
> 
> We never really identified, as in the Demo version we don't "reset" the Demo as often. So, we haven't actually tried to re-seed Demo for a while. Trying to do this right now, yielded the error of the linked ticket.

Okay,  got it. Thanks

beesaferoot (approved):

Issue #84, opened 2024-05-30
Download buttons for the ‘Customers’ and ‘Transactions’ sections


dmohns:
self-explanatory

munyanezaarmel:
@dmohns  In what format will the downloaded files be, and how will the downloaded data be presented?

dmohns:
> [@dmohns](https://github.com/dmohns) In what format will the downloaded files be, and how will the downloaded data be presented?

I think we can start with a very simple CSV export. I.e. just add an export to the Customers and Transactions List pages which download all transactions in CSV.

PR #676, opened 2025-04-17
Add export functionality for customers and transactions

beesaferoot:

Endpoint is needed for #84

### Pull Request checklist


src/backend/routes/api.php
```diff
@@ -321,6 +321,7 @@ static function () {
 Route::group(['prefix' => 'export'], static function () {
     Route::get('/transactions/{slug}', 'TransactionExportController@download');
     Route::get('/debts/{slug}', 'OutstandingDebtsExportController@download');
+    Route::get('/customers/{slug}', 'PersonExportController@download');
```
dmohns, line 324: Could you explain what this `slug` is? In the collection above I see it's set to email of the user 🤔 Why is this needed? 
Looking at the other exports it "looks like" this was meant to be a way to pass a filter for the export, but it doesn't appear to be functional

src/backend/app/Services/AbstractExportService.php
```diff
@@ -117,4 +117,20 @@ public function saveSpreadSheet(): string {
...
+
+            // Create CSV writer
+            $writer = IOFactory::createWriter($this->spreadsheet, 'Csv');
+            $writer->save($fileName);
+
+            return $fileName;
+        } catch (\Exception $e) {
+            throw new SpreadSheetNotSavedException($e->getMessage());
+        }
+    }
```
dmohns, line 135: Do I understand correctly, to create a simple CSV file, we are 

- Creating a spreadsheet by reading a xlsx template
- Transforming our data and inserting it into specific cell in the xlsx
- Convert xlsx to csv

Is this really the way to do it 😅 ? I wonder if there is a more simple way in PHP to achieve this.

collections/Export/Export Transactions Data CSV.bru
```diff
@@ -0,0 +1,27 @@
...
+
+body:json {
+  {
+    "deviceType": "meter",
+    "serial_number": "a98c04c2-470c-31c1-8b6a-c32e038b436b",
+    "provider": "wave_money_transaction",
+    "status": 1,
+    "currency": "TSZ",
+    "timeZone": "UTC"
+  }
```
dmohns, line 25: it's a `get`, probably should have a `body` right?

dmohns (commented):
Hey @beesaferoot I'm going to leave a bunch of comments. Some of them might not be directly related to the work you are doing here. If so, let's discuss if we create follow-up issues for them

src/backend/routes/api.php
beesaferoot: I found that the slug is used as an identifier, which is specifically an email (this is some way to generalize exporting records)

See https://github.com/EnAccess/micropowermanager/blob/360eb39783d5536c947060cd29aec1d3a15e347a/src/backend/app/modules/TenantResolver/ApiResolvers/DataExportResolver.php#L19

src/backend/app/Services/AbstractExportService.php
beesaferoot: 😁 yeah I will find a simpler approach to this. Thanks

src/backend/routes/api.php
dmohns: Ufff 😅 Does this mean that anyone just send an HTTP call `/api/export/customers/demo_company_admin@example.com?format=csv` ? I.e. if I can "guess" a companies email, I can export their entire data from MPM?

collections/Export/Export Transactions Data CSV.bru
dmohns: This look like a collection for a `GET` request to `/api/export/transactions/`... Am I understanding correctly that we are using a payload here (i.e. `body` field) ?

If so:

I know it's technically possible to use `body` in a `GET` request. But by convention, we don't use request body in `GET` requests. It's even discouraged in https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1

>   A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.

src/backend/routes/api.php
beesaferoot: Oh that is correct at the moment 😱

The company ID should probably not be resolved by a user-given identifier.  Good catch

collections/Export/Export Transactions Data CSV.bru
beesaferoot: Thanks for clarifying. We can either make it a post request or use query params. 

What do you suggest?

dmohns: I'd say query params would be most intuitive here. As we indeed "retrieving" information.

beesaferoot:
Ready for review @dmohns

src/backend/routes/api.php
```diff
@@ -318,9 +318,10 @@ static function () {
     Route::get('/{serialNumber}', 'EBikeController@show');
     Route::post('/switch', 'EBikeController@switch');
 });
-Route::group(['prefix' => 'export'], static function () {
-    Route::get('/transactions/{slug}', 'TransactionExportController@download');
-    Route::get('/debts/{slug}', 'OutstandingDebtsExportController@download');
+Route::group(['prefix' => 'export', 'middleware' => 'api'], static function () {
+    Route::get('/transactions', 'TransactionExportController@download');
+    Route::get('/debts', 'OutstandingDebtsExportController@download');
```
dmohns, line 323: While I like changing the existing endpoints here as well, this does require a frontend change as well. Do adapt to the new routes.

dmohns (commented):
Just one comment left regarding frontend. We can either do the frontend change on this PR or create a new one for it. Let me know what you prefer.

src/backend/routes/api.php
beesaferoot: Updated the frontend to reflect route changes. Kindly review

dmohns (approved):
Great 🤩 Thanks 🙏

Issue #683, opened 2025-04-24
[Bug]: `Illuminate\Contracts\Encryption\DecryptException` The payload is invalid. When opening any page with password protection.


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Expected Behavior

Should be able to open anypage with password protection.

### Actual Behavior

`The payload is invalid.`

### Steps to reproduce

Open any page with password protection.

### What operating system(s) are you using?

macOS

PR #684, opened 2025-04-24
Fix `decrypter` error

dmohns:

It turns out that https://github.com/EnAccess/micropowermanager/pull/659 has some massive side effects. Namely, it will delete the "Protected Pages Password" from each Tenant.

Furthermore if the migration is run with existing Tenant data being present, opening pages with password protection will result in an error, because `decrypt` is called on `null`.

This PR addresses this by setting the "Protected Pages Password" to `123123`.

Closes: #683 

Yes, existing Protected Pages Password will be set to `123123`.

Locally.

### Alternatives considered

The alternative would be to revert https://github.com/EnAccess/micropowermanager/pull/659 and try to find a non-breaking solution. Regardless these change were already released in `v0.0.31` so anyone who applied `v0.0.31` would already have their Protected Pages Password deleted already. 

So even if we reverted, we would need to find a solution for both cases, i.e. people coming from `v0.0.30` and `v0.0.31` which might be overly complicated.

### Pull Request checklist

### DIFF
--- a/.github/workflows/erd-generate.yaml
+++ b/.github/workflows/erd-generate.yaml
@@ -38,6 +38,11 @@ jobs:
           extensions: mbstring, dom, fileinfo, mysql
           coverage: xdebug
 
+      - name: Config environment file
+        uses: c-py/action-dotenv-to-setenv@v5
+        with:
+          env-file: ./dev/.env.micropowermanager-ci
+
       - name: Install Composer Dependencies
         run: |
           composer install --no-interaction --prefer-dist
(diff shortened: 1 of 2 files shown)


src/backend/app/Services/CompanyService.php
```diff
@@ -25,11 +24,6 @@ public function getByDatabaseProxy($databaseProxy): Company {
 
     public function getById($id): Company {
         $result = $this->company->newQuery()->findOrFail($id);
-        if (isset($result->protected_page_password)) {
-            if (str_starts_with($result->protected_page_password, 'eyJ')) {
```
beesaferoot, line 29: Why are we removing this? This should not be affected even if `protected_page_password` is null.

dmohns: Yeah, it was just a leftover I think 🤔 We can continue this when we bring back the feature in a new PR.

Issue #493, opened 2025-02-19
[Feature Request]: Display SHS Tokens in the application


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

This is similar but distinct from: #396

Currently, there isn't an easy way to see tokens for **SHS** in the application.

### Proposed Solution

SHS Tokens should be displayed in a similar way to how energy tokens are shown.

### Alternatives Considered

N/A

### Additional Information

Potential additional tasks:

- Currently our Demo data seeding doesn't contain SHS token. We would need to add a few SHS Tokens to the Demo data seed for this to make sense.

beesaferoot:
Hi @dmohns I need more info on this. Where are SHS tokens stored, and is this currently accessible, or do we plan to implement the feature?

dmohns:
Hey @beesaferoot 

Unfortunately, this issue does contain a little bit of research work, as it's not fully clear to me either.

Giving you some hints here:

- In the INENSUS version of MPM I was able to see some SHS Token being displayed in the "Log" section of SoldAppliances. This seems to be done in the Manufacturer implementations, for example [here](https://github.com/EnAccess/micropowermanager/blob/main/src/backend/packages/inensus/sun-king-shs/src/Modules/Api/SunKingSHSApi.php#L92-L106) and [here](https://github.com/EnAccess/micropowermanager/blob/main/src/backend/packages/inensus/angaza-shs/src/Modules/Api/AngazaSHSApi.php#L64-L75). 
- However, we do seem to have a "unified" interface via the `chargeDevice` method. We need to confirm that this is actually processed by https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/Jobs/TokenProcessor.php#L115-L121 (which from what I understand stores an SHS Token)

So the first step would really be to find out, if we are using the same `Token` object of SHS/SoldAppliances as we are using for Electricity Meters.

After that there can be multiple outputs here:

[...]

PR #704, opened 2025-05-08
Support Solar Home System tokens 

beesaferoot:

This PR creates a mechanism for tracking SHS tokens, making it possible to show SHS-specific tokens.

initial work on #493

### Pull Request checklist

### DIFF
--- a/src/backend/app/Jobs/TokenProcessor.php
+++ b/src/backend/app/Jobs/TokenProcessor.php
@@ -4,6 +4,7 @@ namespace App\Jobs;
 
 use App\Misc\TransactionDataContainer;
 use App\Models\AssetRate;
+use App\Models\Meter\Meter;
 use App\Models\Token;
 use Illuminate\Bus\Queueable;
 use Illuminate\Foundation\Bus\Dispatchable;
--- a/src/backend/app/Jobs/TokenProcessor.php
+++ b/src/backend/app/Jobs/TokenProcessor.php
@@ -4,7 +4,6 @@ namespace App\Jobs;
 
 use App\Misc\TransactionDataContainer;
 use App\Models\AssetRate;
-use App\Models\Meter\Meter;
 use App\Models\Token;
 use Illuminate\Bus\Queueable;
 use Illuminate\Foundation\Bus\Dispatchable;
--- a/src/backend/app/Misc/TransactionDataContainer.php
+++ b/src/backend/app/Misc/TransactionDataContainer.php
@@ -7,6 +7,7 @@ use App\Models\Device;
 use App\Models\Manufacturer;
 use App\Models\Meter\Meter;
 use App\Models\Meter\MeterTariff;
+use App\Models\SolarHomeSystem;
 use App\Models\Token;
 use App\Models\Transaction\Transaction;
 use App\Services\AppliancePaymentService;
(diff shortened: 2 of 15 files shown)


dmohns (commented):
I like this a lot!! Thanks! 

Just one thought I had about the SHSToken..: In the SHSToken you are distinguishing between `TYPE_TIME` and `TYPE_ENERGY`. While it is imaginable to use SHS with Energy based tokens, I think the majority of use-cases would be time based. I.e. "enable the SHS for 1 Month".
At the same time I'm wondering if SHSToken is to particular. Say for example we want to add Time Based tokens for other appliances in the future, then it would be odd to use `Token` because it's for Meter, and SHSToken because it's for SHS.

What do you think about just using `Token` for both? I.e. add all the additional fields you are adding for SHSToken to Token directly (especially the `device` relation makes so much sense). And the additional `token_type` field?

beesaferoot:
> I like this a lot!! Thanks!
> 
> Just one thought I had about the SHSToken..: In the SHSToken you are distinguishing between `TYPE_TIME` and `TYPE_ENERGY`. While it is imaginable to use SHS with Energy based tokens, I think the majority of use-cases would be time based. I.e. "enable the SHS for 1 Month". At the same time I'm wondering if SHSToken is to particular. Say for example we want to add Time Based tokens for other appliances in the future, then it would be odd to use `Token` because it's for Meter, and SHSToken because it's for SHS.
> 
> What do you think about just using `Token` for both? I.e. add all the additional fields you are adding for SHSToken to Token directly (especially the `device` relation makes so much sense). And the additional `token_type` field?
> 
> Alternatively, if we decide to go with the SHSToken as you suggest in the current PR, wdyt about renaming the current `Token` to `MeterToken` ?


Thanks for the review. 🙌🏾

The reason for having SHSTokens is to allow tracking SHS-specific tokens separate from the general tokens (i.e Tokens).  This is similar to how we have  `Token` but still have `MeterToken`  

Here we have both `MeterToken` and `SHSToken` which are all referenced in `Tokens` record, because there are both created from Tokens. 

Pertaining to the energy/time based separation, I will refactor to only account for time based. Initially thought this might matter after checking the two SHS custom plugins.  There is no reason to assume that a future manufacturer may not go with the energy based version. 

```php
    public function chargeDevice(TransactionDataContainer $transactionContainer): array {
        $dayDifferenceBetweenTwoInstallments = $transactionContainer->dayDifferenceBetweenTwoInstallments;
        $minimumPurchaseAmount = $transactionContainer->installmentCost;
        $minimumPurchaseAmountPerDay = ($minimumPurchaseAmount / $dayDifferenceBetweenTwoInstallments); // This is for 1 day of energy
        $transactionContainer->chargedEnergy = 0; // will represent the day count
        $transactionContainer->chargedEnergy += ceil($transactionContainer->rawAmount / $minimumPurchaseAmountPerDay);

        Log::debug('ENERGY TO BE CHARGED as Day '.$transactionContainer->chargedEnergy.
            ' Manufacturer => SunKingSHSApi');

        $energy = $transactionContainer->chargedEnergy;
        $params = $this->buildParams($transactionContainer);
        $credentials = $this->credentialService->getCredentials();
        $response = $this->handleApiRequest($credentials, $params);

        $this->recordTransaction($transactionContainer);
        $this->storeToken($transactionContainer, $response['token'], $energy);
        $this->logAction($transactionContainer, $response['token']);

        return [
            'token' => $response['token'],
            'load' => $energy,
        ];
    }
 ```

beesaferoot:
> > The reason for having SHSTokens is to allow tracking SHS-specific tokens separate from the general tokens (i.e Tokens). This is similar to how we have `Token` but still have `MeterToken`
> > Here we have both `MeterToken` and `SHSToken` which are all referenced in `Tokens` record, because there are both created from Tokens.
> 
> Waaaaiit.. we already have `MeterToken` 😅 ? My apologies I wasn't aware of this. This seems weird. I cannot seem to find any relation between `Token` and `MeterToken`, am I missing something?

Well, not a DB relationship, but more so the tokens in MeterToken are also present in Token.

beesaferoot:
Looking at the data, there seems to be a disparity on the tokens in the `Token` and `MeterToken` even there have the same transaction id. I think this might be caused from the Seeding logic. I will investigate

Screenshot 2025-05-12 at 10 06 03 AM

Screenshot 2025-05-12 at 10 07 12 AM

dmohns:
Alright, so the two path I can see here:

1. Add the new `SHSToken` and fix the existing `Token<>MeterToken` relation. I.e. the token value's should not be stored at `Token` level.
2. Remove the `MeterToken` and store all information at `Token` level. Then add a distinction there along the lines of `TOKEN_TYPE`.

dmohns (approved):
Nice!! Thanks

PR #711, opened 2025-05-12
Refactor Token handling by consolidating MeterToken into Token model

### LINKED ISSUE #709: [Feature Request]:  Consolidate MeterToken into generic Token
### Preflight Checklist

### Problem Description

Following a discussion on #704, it will be a better approach to consolidate MeterToken into Token to further reduce the complexity of the system. 

### Proposed Solution

Remove `MeterToken` entirely in favour of a more generic `Token` model.   This setups the work for expanding the token support to allow shs tokens. 

### Alternatives Considered

Multiple altervatives were considered.

### Additional Information

_No response_

beesaferoot:

Closes #709 
### Pull Request checklist


src/backend/app/Models/Token.php
```diff
@@ -17,4 +56,11 @@ public function transaction(): BelongsTo {
     public function paymentHistories(): MorphOne {
         return $this->morphOne(PaymentHistory::class, 'paid_for');
     }
+
+    public function soldEnergyPerPeriod($startDate, $endDate): Builder {
+        return $this::query()
+            ->select(DB::raw(' SUM(load) as sold,YEARWEEK(created_at,3) as period'))
+            ->whereBetween('created_at', [$startDate, $endDate])
+            ->groupBy(DB::raw('YEARWEEK(created_at,3)'));
+    }
```
dmohns, line 65: This would only make sense for `kWH` Tokens, so we should add a filter here. I.e. `WHERE token_type = ...`.

However, are we even using this function at all?

src/backend/app/Models/Token.php
```diff
@@ -4,11 +4,50 @@
...
 
+/**
+ * Generic Token Model for all types of tokens i.e meter token, shs token, etc.
+ *
+ * @property string $token
+ * @property float  $load
+ * @property int    $transaction_id
+ * @property string $token_type
+ * @property int    $device_id
+ * @property int    $token_amount
```
dmohns, line 20: `load` and `token_amount` seem redundant.

I like what you had suggested in your previous PR to have the following fields

- `token_type` (can be `time` or `energy`)
- `token_value` (can be `kWH`, `day`, `month`) Although this name is not fully intuitive. Maybe something like `token_unit`?
- `token_amount` (can be the actual amount) TBC if this should better be a float to enable for example 2.5 kWH tokens?

dmohns (commented):
This is great. Few comments

src/backend/app/Models/Token.php
beesaferoot: Alright, I will work on it. Thanks

src/backend/app/Models/Token.php
beesaferoot: > This would only make sense for kWH Tokens, so we should add a filter here. I.e. WHERE token_type = ....

Good catch on this, will fix. 

> However, are we even using this function at all?

I will have to check, it existed for MeterToken so I just ported it.

src/backend/app/Models/Token.php
beesaferoot: How do you suggest getting rid of an already existing `load` table field? I didn't want to potential break existing company table.

dmohns: Can't we just rename it in the migration? Then, of course, we would have to check where in the code `MeterToken->load` is used, and replace it with the new field name.

But maybe I'm thinking to easy 😅

beesaferoot: 😄 easy/simple solutions are always the best. I will go with this; I was overthinking it 🙃

src/backend/app/Jobs/TokenProcessor.php
```diff
@@ -113,7 +113,12 @@ private function retryTokenGeneration(): void {
     }
 
     private function saveToken(array $tokenData): void {
-        $token = Token::query()->make(['token' => $tokenData['token'], 'load' => $tokenData['load']]);
+        $token = Token::query()->make(['token' => $tokenData['token'], 'token_amount' => $tokenData['load']]);
+        $token->device_id = $this->transactionContainer->device->id;
+        if ($this->transactionContainer->meter !== null) {
+            $token->token_type = Token::TYPE_ENERGY;
+            $token->token_unit = Token::UNIT_KWH;
+        }
```
dmohns, line 121: From what I understand the way `saveToken` is used is that it takes the output of the various manufacturer's `chargeDevice`.

The interface is already quite odd... For example AngazaSHS returns

```
        return [
            'token' => $token,
            'load' => $energy,
        ];
```

But here `load` is actually "days".

Should also demand the new-style token here? I.e. require `$tokenData` to be passed like this

```php
[
        'token' => $token,
        'token_type' => ...,
        'token_unit' => ...,
        'token_amount' => ...,
];
```

This would require changing all `chargeDevice` functions of all Providers, but would be a great change.

src/backend/app/Models/Meter/Meter.php
```diff
@@ -64,8 +65,9 @@ public function accessRate(): AccessRate {
         return $this->tariff->accessRate;
     }
 
-    public function tokens(): HasMany {
-        return $this->hasMany(MeterToken::class);
+    public function tokens(): \Illuminate\Database\Eloquent\Relations\HasManyThrough {
+        return $this->hasManyThrough(Token::class, Device::class, 'device_id', 'device_id', 'id', 'id')
+            ->where('device_type', Meter::class);
```
dmohns, line 70: This is great, I wasn't aware we could do this `hasManyThrough`!!! Should we also add this to SHS model?

dmohns (commented):
Really great. Few more comments.

src/backend/app/Models/Meter/Meter.php
beesaferoot: Will add this in #704

PR #720, opened 2025-05-22
Enable customers (app) list pagination

### LINKED ISSUE #719: [Bug]: Agent App should support pagination
### Preflight Checklist

### Expected Behavior

Scrolling down the customers list should fetch more data via pagination. 

### Actual Behavior

App throws an error due to backend responses. 

### Steps to reproduce

Work through;
1. Go to customers section
2. Open list of customers 
3. Scroll to buttom

macOS

### Additional Information

TBD

beesaferoot:

Closes #719 

### Pull Request checklist

### DIFF
--- a/src/backend/app/Http/Middleware/TrustProxies.php
+++ b/src/backend/app/Http/Middleware/TrustProxies.php
@@ -11,12 +11,7 @@ class TrustProxies extends Middleware {
      *
      * @var array<int, string>|string|null
      */
-    protected $proxies = [
-        '127.0.0.1',
-        '::1',
-        // Add your production proxy IPs here
-        // For example: '10.0.0.0/8', '172.16.0.0/12', '192.168.0.0/16'
-    ];
+    protected $proxies;
 
     /**
      * The headers that should be used to detect proxies.
--- a/src/backend/app/Http/Middleware/TrustProxies.php
+++ b/src/backend/app/Http/Middleware/TrustProxies.php
@@ -11,14 +11,18 @@ class TrustProxies extends Middleware {
      *
      * @var array<int, string>|string|null
      */
-    protected $proxies;
+    protected $proxies = [
+        '127.0.0.1',
+        '::1',
+        // Add your production proxy IPs here
+        // For example: '10.0.0.0/8', '172.16.0.0/12', '192.168.0.0/16'
+    ];
 
     /**
      * The headers that should be used to detect proxies.
      *
      * @var int
      */
-    // After...
     protected $headers =
         Request::HEADER_X_FORWARDED_FOR |
         Request::HEADER_X_FORWARDED_HOST |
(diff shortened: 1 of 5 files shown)


beesaferoot:
The app encounters an error when attempting to load the next page of customer data using the next_page_url provided in the paginated server response.

Although the initial request is made over HTTPS, the server returns a next_page_url with the HTTP scheme. This discrepancy causes the mobile client to treat it as a new origin, which results in the request headers—particularly the authorization token—being reset. As a result, the server receives the request without authentication and responds with an error.

src/backend/app/Http/Middleware/TrustProxies.php
```diff
@@ -11,14 +11,18 @@ class TrustProxies extends Middleware {
      *
      * @var array<int, string>|string|null
      */
-    protected $proxies;
+    protected $proxies = [
+        '127.0.0.1',
+        '::1',
+        // Add your production proxy IPs here
+        // For example: '10.0.0.0/8', '172.16.0.0/12', '192.168.0.0/16'
+    ];
```
dmohns, line 19: I don't think this does anything. From what I understand in our set up the Proxy IP's are the ones from GCE LoadBalancers: https://cloud.google.com/load-balancing/docs/https#global-classic-connections

However, I don't think we want to "hardcode" GFE here, else it will be hard for people to re-use the code. Looking around, it seems the Middleware internally uses `trustedproxy.proxies`

```php
$trustedIps = $this->proxies() ?: config('trustedproxy.proxies');
```

So, maybe we can just add a `config` that reads it from ENV.
Example can be found here: https://github.com/laravel/docs/pull/7457/

And then, we can add the GFE IP ranges to `configmap-cloud|demo-backend.yaml`. And also add it to deployment guide, as this is not trivial (I think).

src/backend/app/Providers/AppServiceProvider.php
```diff
@@ -83,6 +84,10 @@ public function boot() {
                 EBike::RELATION_NAME => EBike::class,
             ]
         );
+
+        if (config('app.env') === 'production') {
+            URL::forceScheme('https');
+        }
```
dmohns, line 90: I wonder if this is too narrow, and could cause some problems down the line. It should not be required if the Trust proxies are set up correctly, see above.

dmohns (commented):
Wow, great find! Left some comments on your solution.

src/backend/app/Http/Middleware/TrustProxies.php
beesaferoot: Thanks for the explanation, I will be looking in this.

beesaferoot: Thanks for the explanation, I will be looking in this.

PR #781, opened 2025-07-08
Add protected pages password documentation

dmohns:
### Pull Request checklist


dmohns:
Ohh, really good catches!

Regarding 1. and 3.:

I had the feeling I'm missing something, this was it. I still need to update `protected_pages` table via a migration file to change the new route locations. Will add that 👍 

Regarding 2.:

Yeah, I forgot to remove the button itself 🤦 I only removed the modal, so the button doesn't do anything right now. Will fix.

beesaferoot:
Thanks for addressing them. I have one last concern. 

Now the user management page is protected, which means you need to know the password before you can change it 😄

In Aaron's case, i.e., forgetting the password, it will not be possible to change it with what we have now. 


Maybe for future improvement we need a way to change one's PPP (via email?) without having to remember it.

dmohns:
> Thanks for addressing them. I have one last concern.
> 
> Now the user management page is protected, which means you need to know the password before you can change it 😄
> 
> In Aaron's case, i.e., forgetting the password, it will not be possible to change it with what we have now.
> 
> Maybe for future improvement we need a way to change one's PPP (via email?) without having to remember it.

Hmm... good point! 

But thinking about this... We should also consider the reverse 😅: In the current way **everyone** can change the Protected Pages password?! Basically rendering it use-less, no 🫣?.

I think it's just a limitation of the (poor) design of the PPP as a security feature.

beesaferoot (dismissed):

PR #817, opened 2025-07-20
Fix cluster dashboard revenue service

beesaferoot:
This fixes a regression on the dashboard page following recent larastan refactoring.

### Pull Request checklist

### DIFF
--- a/src/backend/app/Services/ClusterRevenueService.php
+++ b/src/backend/app/Services/ClusterRevenueService.php
@@ -114,6 +114,15 @@ class ClusterRevenueService {
             ->groupBy(DB::raw('DATE_FORMAT(created_at,\'%Y-%m\'),WEEKOFYEAR(created_at)'))->get();
     }
 
+    /**
+     * @param Collection<int, Cluster> $clusters
+     * @param string                   $startDate
+     * @param string                   $endDate
+     * @param array                    $periods
+     * @param string                   $period
+     *
+     * @return array
+     */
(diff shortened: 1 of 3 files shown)


src/backend/app/Services/ClusterRevenueService.php
```diff
@@ -82,7 +82,7 @@ static function ($q) {
      */
     public function getTransactionsForWeeklyPeriod(int $clusterId, array $period, ?int $connectionType = null): Collection {
         return $this->transaction->newQuery()
-            ->selectRaw('DATE_FORMAT(created_at,\'%Y-%m\') as period , SUM(amount) as revenue')
+            ->selectRaw('DATE_FORMAT(created_at,\'%Y-%m\') as period , SUM(amount) as revenue, WEEKOFYEAR(created_at) as week')
```
dmohns, line 85: I don't understand why this wasn't necessary before Larastan fixes 😂 But @alaincodes24 suggested the same change.

dmohns (approved):

src/backend/app/Services/ClusterRevenueService.php
beesaferoot: The error seems to have been suppressed because when you convert the transaction object to an array, you get only the fields `[ period => '', revenue => ''`, indexing that array by "week" throws an error.

We starting converting the revenueTransaciton returned db object to an array after the Larastan fixes.

PR #818, opened 2025-07-20
refactor: use correct error response message reference (data.message)

beesaferoot:
Fixes undefined reference to error response message value `response.data.data.message` to valid  message reference `response.data.message`)

### Pull Request checklist


beesaferoot:
> Thanks! Any idea how this error happened? I.e. was this always wrong or did some refactoring introduced it?

Unsure when it was introduced, but I noticed we had several references to `response.data.message` already, so my guess is we used the wrong reference earlier refactoring.

PR #820, opened 2025-07-21
Add install commands in app run on the web

### LINKED ISSUE #812: [Bug]: First time activating a plugin fails with `The command "<plugin>:install" does not exist.`
### Preflight Checklist

### Expected Behavior

When enabling a plugin for the first time (either via Company registration or via settings) MPM throws an error `The command "<plugin>:install" does not exist.`.

Consecutive attempts do succeed.

However, the install commands are not actually executed. A good example is the following: If for example the SparkMeter plugin is enabled, Spark should be added to the manufacturers table, which doesn't happen.

### Actual Behavior

see above

### Steps to reproduce

see above

macOS

### Additional Information

_No response_

beesaferoot:
Closes #812

### Pull Request checklist

### DIFF
--- a/src/backend/bootstrap/app.php
+++ b/src/backend/bootstrap/app.php
@@ -78,5 +78,4 @@ return Application::configure(basePath: dirname(__DIR__))
         // will run on the last day of the month
         $schedule->command(MailApplianceDebtsCommand::class)->weeklyOn(1, '6:00');
     })
-    ->withProviders(require __DIR__.'/providers.php')
     ->create();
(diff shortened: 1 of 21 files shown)


src/backend/bootstrap/app.php
```diff
@@ -78,4 +78,5 @@
         // will run on the last day of the month
         $schedule->command(MailApplianceDebtsCommand::class)->weeklyOn(1, '6:00');
     })
+    ->withProviders(require __DIR__.'/providers.php')
```
dmohns, line 81: How did you find these? I'm reading the documentation and I cannot find any trace of it saying that we need to specifically load `providers.php` 😖

dmohns: The few examples, that I do find use 

```
->withProviders()
```

Would this work for us as well? I think our bootstrap file is in the default location.

beesaferoot: Removed the method and it still works as expected.

dmohns (approved):

src/backend/bootstrap/app.php
beesaferoot: True, a bunch of plugins we no longer support are still hanging around.

PR #826, opened 2025-07-24
Publish Queue Worker containers to Dockerhub

dmohns:

Goal of this PR is to enable the queue workers for job processing on our cloud instances.

- [x] Make the Horizon Dashboard accessible
- [x] Simplify Queue and Connection management
- [x] Publish Queue Worker Docker containers
- [x] Update deployment to use new containers

**Some noteworthy improvements**

- Offloading all package directories `vendor` and `node_modules` to dedicated volumes now in local development. This fixes an annoying race condition when multiple containers use the `vendor` directory from the host.
- Removed unneeded volume mounts from Dockerfiles in prod mode.
- Moved `Connection` and `Queue` definitions to the Job. Simplifying all Job dispatching code from
  ```php
  ProcessPayment::dispatch($transaction->id)
                  ->allOnConnection('redis')
                  ->onQueue(config('services.queues.payment'));
  ```
  to just
  ```php
  ProcessPayment::dispatch($transaction->id);
  ```
- Removing possibility to customise Queue names via Environment variables (the queue names are already hardcoded in the Job definitions. If users change the queue names via env vars the jobs will not get picked up).
- Remove support for the `database` queue and drop the no-longer needed `jobs` table. We don't even have a queue worker for it. Moving forward we will run all jobs on Redis. However, the default with stay `sync` in-case something quick has to be run synchronously. (Note: We are still keeping `company_jobs` for now. Separate clean up request here: https://github.com/EnAccess/micropowermanager/issues/837)

### Pull Request checklist


beesaferoot:
Hi @dmohns, Excellent work getting this up. Just to get a better idea what this accomplishes, can you try explaining why th e previous approach with `QUEUE_DRIVER` didn't work and the move to this? 

Apologies if the questions are obvious. 😄

beesaferoot:
> > Hi @dmohns, Excellent work getting this up. Just to get a better idea what this accomplishes, can you try explaining why th e previous approach with `QUEUE_DRIVER` didn't work and the move to this?
> > Apologies if the questions are obvious. 😄
> 
> Hey @beesaferoot
> 
> No worries, it's not obvious at all! It took me many hours to get to the bottom of this 😅
> 
> * The queue **did** work before, but only "kinda". Our default queue driver was set to `database`. This queue did not work, because we don't have any queue worker for this connection. (via `artisan queue:work`). Our worker worker uses Horizon which only works on the Redis queue connection. However, all Job Dispatch calls would actually customise the connection to use Redis.
> * If you were to write a new Job and forgot to specify the connection it would not run. With this refactor it would at least run synchronously.
> * The Horizon queue worker was not getting deployed on k8s yet. I'm updating the Dockerfiles to better match our standard.
> 
> In the end the main goal of all of what I'm doing here, is to make things easier to understand (I hope).

This really helps to clarify things. Thank you @dmohns

beesaferoot (approved):

PR #827, opened 2025-07-26
Add persistent volume claims for storage and logs in EKS and GKE deployments

### LINKED ISSUE #599: [Bug]: Generated report not persistent upon pod restarts
### Preflight Checklist

### Expected Behavior

When MicroPowerManager is ran in a Kubernetes cluster a pod restart should not negatively affect the MicroPowerManager functionality.

### Actual Behavior

When we restart a pod all previously generated reports are lost. This is because reports are persisted to disk.

### Steps to reproduce

Generate a report and restart the pod.

macOS

### Additional Information

Multiple solutions may exists. Either we 

- change our k8s setup to add a persistent storage to location where we store reports (and potentially other things)
- change the way reports are generated, such they are persisted not on disk, but elsewhere (database, cloud storage, etc...)

beesaferoot:
Closes #599 

### Pull Request checklist

### DIFF
--- a/k8s/base/aws_eks/deployments/mpm-backend.yaml
+++ b/k8s/base/aws_eks/deployments/mpm-backend.yaml
@@ -23,3 +23,17 @@ spec:
                 name: mpm-configmap-backend
             - secretRef:
                 name: mpm-secret-backend
+          volumeMounts:
+            - name: storage-data
+              mountPath: /var/www/html/storage
+              subPath: ""
+            - name: logs-data
+              mountPath: /var/www/html/storage/logs
+              subPath: "logs"
+      volumes:
+        - name: storage-data
+          persistentVolumeClaim:
+            claimName: mpm-storage-pvc
+        - name: logs-data
+          persistentVolumeClaim:
+            claimName: mpm-logs-pvc
--- a/k8s/base/gcp_gke/deployments.yaml
+++ b/k8s/base/gcp_gke/deployments.yaml
@@ -23,6 +23,20 @@ spec:
                 name: mpm-configmap-backend
             - secretRef:
                 name: mpm-secret-backend
+          volumeMounts:
+            - name: storage-data
+              mountPath: /var/www/html/storage
+              subPath: ""
+            - name: logs-data
+              mountPath: /var/www/html/storage/logs
+              subPath: "logs"
+      volumes:
+        - name: storage-data
+          persistentVolumeClaim:
+            claimName: mpm-storage-pvc
+        - name: logs-data
+          persistentVolumeClaim:
+            claimName: mpm-logs-pvc
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -74,3 +88,17 @@ spec:
                 name: mpm-configmap-backend
             - secretRef:
                 name: mpm-secret-backend
+          volumeMounts:
+            - name: storage-data
+              mountPath: /var/www/html/storage
+              subPath: ""
+            - name: logs-data
+              mountPath: /var/www/html/storage/logs
+              subPath: "logs"
+      volumes:
+        - name: storage-data
+          persistentVolumeClaim:
+            claimName: mpm-storage-pvc
+        - name: logs-data
+          persistentVolumeClaim:
+            claimName: mpm-logs-pvc
--- a/k8s/base/gcp_gke/kustomization.yaml
+++ b/k8s/base/gcp_gke/kustomization.yaml
@@ -5,6 +5,7 @@ resources:
   - configmaps.yaml
   - redis.yaml
   - deployments.yaml
+  - mpm-storage-pvc.yaml
   - services.yaml
   - ingress.yaml
   - egress-nat-policy.yaml
--- a/k8s/base/aws_eks/deployments/mpm-scheduler.yaml
+++ b/k8s/base/aws_eks/deployments/mpm-scheduler.yaml
@@ -22,3 +22,17 @@ spec:
                 name: mpm-configmap-backend
             - secretRef:
                 name: mpm-secret-backend
+          volumeMounts:
+            - name: storage-data
+              mountPath: /var/www/html/storage
+              subPath: ""
+            - name: logs-data
+              mountPath: /var/www/html/storage/logs
+              subPath: "logs"
+      volumes:
+        - name: storage-data
+          persistentVolumeClaim:
+            claimName: mpm-storage-pvc
+        - name: logs-data
+          persistentVolumeClaim:
+            claimName: mpm-logs-pvc
--- a/k8s/base/aws_eks/deployments/mpm-storage-pvc.yaml
+++ b/k8s/base/aws_eks/deployments/mpm-storage-pvc.yaml
@@ -1,0 +1,23 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: mpm-storage-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: gp2  # AWS EBS storage class
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: mpm-logs-pvc
+spec:
+  accessModes:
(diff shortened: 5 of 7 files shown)


k8s/base/gcp_gke/kustomization.yaml
```diff
@@ -5,6 +5,7 @@ resources:
   - configmaps.yaml
   - redis.yaml
   - deployments.yaml
+  - mpm-storage-pvc.yaml
```
dmohns, line 8: Can we rename the file to skip the `mpm-` prefix here? Just to be consistent with other files in this folder.

k8s/base/aws_eks/deployments/mpm-backend.yaml
```diff
@@ -23,3 +23,17 @@ spec:
                 name: mpm-configmap-backend
             - secretRef:
                 name: mpm-secret-backend
+          volumeMounts:
+            - name: storage-data
+              mountPath: /var/www/html/storage
+              subPath: ""
+            - name: logs-data
+              mountPath: /var/www/html/storage/logs
+              subPath: logs
```
dmohns, line 32: I don't think we need a separate `logs` mount here. After all, we are not even using logging to file anymore. It was moved to STDERR/STDOUT logging which works much better with `k8s`.

(applies to all files in this PR)

k8s/base/gcp_gke/deployments.yaml
```diff
@@ -23,6 +23,20 @@ spec:
...
+            - name: storage-data
+              mountPath: /var/www/html/storage
+              subPath: ""
+            - name: logs-data
+              mountPath: /var/www/html/storage/logs
+              subPath: logs
+      volumes:
+        - name: storage-data
+          persistentVolumeClaim:
+            claimName: mpm-storage-pvc
```
dmohns, line 36: I have actually never done this. But will this be a problem if a single PVC is shared between pods?

dmohns (commented):
Left some comments in the code. Also, one more high-level thought:

I want our deployment scenario for prod and the local development environment be relatively similar. It feels a bit random that we are suddenly mounting a PVC in (and only in) prod deployment. To make this a bit more traceable what do you think about the following (additional) steps?

1. Add `VOLUME ["/var/www/html/storage"]` to all our Dockerfiles (to indicate that this folder should be persisted)
2. Add an actual Volume to our local setup. Similar to how we persist data for `mysql_data` and `redis_data` locally

k8s/base/gcp_gke/deployments.yaml
beesaferoot: From the k8s docs multiple pods can share a single PVC with the right access mode. I used `ReadWriteOnce` which allows pods from "the same" node read and write to the volume. 

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

k8s/base/aws_eks/deployments/mpm-backend.yaml
beesaferoot: Noted.

dmohns:
Hmmmm 🙈 I'm afraid this one also, might not be so easy 😹 

It looks like we are using the Storage Directory a little bit wrong. It is meant for runtime generated files only, see [here](
https://laravel.com/docs/12.x/structure#the-storage-directory). In which case putting it into a PVC makes total sense.

However, as of right now, we have some persistent in `storage` which MPM requires to function. For example, `countries.json`, `currencies.json`, and some template files. If we now mount a volume into `storage` it will be empty, and the files won't be available.

I think as a previous step we would first need to move all these files out of `storage` (maybe into `resources`?) WDYT?

beesaferoot:
> It looks like we are using the Storage Directory a little bit wrong. It is meant for runtime generated files only, see [here](https://laravel.com/docs/12.x/structure#the-storage-directory). In which case putting it into a PVC makes total sense.

Not entirely runtime-generated as noted below from the docs.  Some static files can be place in the `app/public` directory which I believe is what we have (i.e countries.json) and such. 

> The storage/app/public directory may be used to store user-generated files, such as profile avatars, that should be publicly accessible. You should create a symbolic link at public/storage which points to this directory. You may create the link using the php artisan storage:link Artisan command.


Looking at the access patterns, moving everything from storage would not be ideal. I'm thinking of moving the conventional things that should be in `resources` and then only mount a specific folder inside storage i.e `persisted` that we resolve in the application when we need to persist. 

The volume may now be mounted pointing to that folder;

```yaml 
volumes:
      - ./src/backend:/var/www/html
      -backend_php_vendor:/var/www/html/vendor
      - backend_php_persisted:/var/www/html/storage/persisted/
```

WDYT?

dmohns:
> Looking at the access patterns, moving everything from storage would not be ideal. I'm thinking of moving the conventional things that should be in `resources` and then only mount a specific folder inside storage i.e `persisted` that we resolve in the application when we need to persist.

I was considering something similar before. But I don't like it 🫣 When doing this, we are creating complexity to a problem that is not that complicated. I.e. some folders in `storage` will be persisted, some not, some contain version controlled files, some not.

Ultimately, we want to things

1. A place to store static files, version controlled which MPM needs to function
2. A place to store runtime generated files, not version controlled.

The way I like to think about 2., is that the directory should be empty on the initial start of an MPM instance.

> > It looks like we are using the Storage Directory a little bit wrong. It is meant for runtime generated files only, see [here](https://laravel.com/docs/12.x/structure#the-storage-directory). In which case putting it into a PVC makes total sense.
> 
> Not entirely runtime-generated as noted below from the docs. Some static files can be placed in the `app/public` directory which I believe is what we have (i.e countries.json) and such.

I'm re-reading it again, but its quite clear that all files should be **generated**. They do distinguish whether or not the files can be publicly accessible, but the example they give about avatars doesn't apply to us at all.

If we need `countries.json` (which is static) to be publicly accessible I'd say this should actually go to `public` directory directly.

**Now:** I understand that this task is growing a bit out of scope (you are trying to persist report files). What we could do, is use your suggested approach with a `persisted` directory as an intermediate step. And then tackle the entire `storage` directory in a separate PR. I.e.

1. Create a `persisted` directory in `storage` and mount a PVC to it (this PR).
2. Move all static files out of `storage`
3. Mount the PVC to entire `storage`

WDYT?

dmohns (commented):
Great stuff! I will review this in-depth next week 🕵️

dmohns (commented):
This is great 🤩 Only comment regarding the `storage` folder. Can't we make it entirely empty now? I think I would keep the plain `storage` folder to indicate we are using it, but delete any subfolders in it.

dmohns (commented):
Left some comments in the code. Also more another comment:

- I noticed some files are getting removed. Namely XLSX template files. For example `export_people_template.xlsx` but this seems to be used `PersonExportService`

src/backend/app/Providers/AppServiceProvider.php
```diff
@@ -86,6 +87,22 @@ public function boot() {
...
+
+        // Ensure required storage subdirectories exist during app startup
+        $requiredStorageDirectories = [
+            storage_path('logs'),
+            storage_path('app'),
+            storage_path('app/certs/companies'),
+            storage_path('app/public'),
+            storage_path('outsourcing'),
+            storage_path('appliance'),
+        ];
```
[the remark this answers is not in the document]
beesaferoot: > WDYT about moving the responsibility of creating the required sub-folder structure to corresponding piece? For example, making storage_path('outsourcing') exists could be run in OutsourceReportGenerator etc..?

That works as well. I went with the approach of having it run on startup to ensure the storage location can be written into and is in the expected state before the app is fully initialized. 


> Also, from what I understand Laravel takes care of generating it's own files, like logs and app. Did you encounter any issues when the folders weren't present?

I ran into a permission issue with laravel writing into the storage volume while working on this. But unclear if it had anything to with sub-folders not existing.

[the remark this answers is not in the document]
beesaferoot: Confirming  `app/public` folders are generated automatically if missing.

beesaferoot:
Ready for your humble review @dmohns

PR #829, opened 2025-07-27
Use Laravel builtin Email functionality to send email

beesaferoot:
Having a email sent to stdout on local/development will help while integrating features using emails.

### Pull Request checklist

### DIFF
--- a/dev/.env.micropowermanager-ci
+++ b/dev/.env.micropowermanager-ci
@@ -17,7 +17,3 @@ DB_USERNAME=root
 DB_PASSWORD=wF9zLp2qRxaS2e
 
 JWT_SECRET=TihQ232bmMermM8HKaLHwzRl5JSuxwqhfEAQKxRHrQqjRUwrsw1FRZhAcLkihEDU
-
-# Mailer configuration for tests
-MAIL_LOG_CHANNEL=stdout
-MAIL_MAILER=log


beesaferoot:
> I wonder if there is a more "Laravel" way of achieving this. For example, I noticed there is a `log` driver for emails, see https://laravel.com/docs/12.x/mail#mail-and-local-development
> 
> Could we somehow utilize this instead of having a piece of code just for local testing?

I will look into this further. Thanks for the feedback.

beesaferoot:
Updated to default to Laravel's bultin Mailer to send emails via log channel.

beesaferoot:
@dmohns This is ready for review.

dev/.env.micropowermanager-backend
```diff
@@ -22,3 +22,6 @@ REDIS_PASSWORD=null
 JWT_SECRET=TihQ232bmMermM8HKaLHwzRl5JSuxwqhfEAQKxRHrQqjRUwrsw1FRZhAcLkihEDU
 
 CACHE_DRIVER=redis
+# Mailer configuration for local development
+MAIL_LOG_CHANNEL=stdout
+MAIL_MAILER=log
```
dmohns, line 27: Nit: Can you add back the empty need line at the end 😅

beesaferoot: > Can we can rid of MailHelperInterface entirely?

From what I understand, having a facade (wrapper) like this is what makes it easy to easily swap things when we need to. For example, during tests we can have a simplified implementation.

Refactoring from PHPMailer was straightforward because of the simple interface. 😄

beesaferoot: I can't think of a good enough argument now, especially when we haven't started considering the context of tests. For now I will get rid of it; we can always bring it back in the future when the use case makes more sense.

beesaferoot: Updated.

dmohns (commented):
Can you also double check this section of the docs: https://micropowermanager.io/installation/configuration-production.html#email ?

At least the "Testing" section no longer applies, right? https://micropowermanager.io/installation/configuration-production.html#testing-email-integration

We also need to make sure we update the MPM Cloud accordingly, as we release this.

dev/.env.micropowermanager-backend
dmohns, line 27: Let make this the default? I.e. by changing `config/mail.php`. Seems to be inline with Laravel default's nowadays: https://github.com/laravel/laravel/blob/11.x/config/mail.php#L17

I would also make `MAIL_LOG_CHANNEL=stdout` the default. I don't even see what other value would make sense here 😂

docs/installation/environment-variables.md
```diff
@@ -124,17 +124,15 @@ Slack
...
+| Environment Variable   | Default                                      | Description                                                   |
+| ---------------------- | -------------------------------------------- | ------------------------------------------------------------- |
+| `MAIL_SMTP_HOST`       | `smtp.mailgun.org`                           | Mail server hostname. For example `smtp.mailserver.com`.      |
+| `MAIL_SMTP_PORT`       | `587`                                        | Mail server port.                                             |
+| `MAIL_SMTP_ENCRYPTION` | `tls`                                        | Mail encryption.                                              |
+| `MAIL_SMTP_USERNAME`   | **Required** (when authentication is needed) | The username used in SMTP authentication.                     |
+| `MAIL_SMTP_PASSWORD`   | **Required** (when authentication is needed) | The password used in SMTP authentication.                     |
```
dmohns, line 133: Maybe we can remove the SMPT ones from here, and instead just link to Laravel documentation: https://laravel.com/docs/12.x/mail#driver-prerequisites ?

dmohns (dismissed):
Looks good 👍 

Unrelated to this PR: While reviewing this I noticed we even have a `mail_settings` table and a UI for it. Both seem unused. Should probably clean this up 🫣

beesaferoot:
> Looks good 👍
> 
> Unrelated to this PR: While reviewing this I noticed we even have a `mail_settings` table and a UI for it. Both seem unused. Should probably clean this up 🫣

I will have a look, possibly create an issue for this specifically.

Issue #837, opened 2025-08-05
[Feature Request]: Deprecate "Company Jobs" abstraction


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

MPM uses a so-called `AbstractJob` abstraction to run it's jobs. The idea is to wrap all jobs into `DatabaseProxyManagerService` to select the correct tenant database. Additionally, keeping a log of all jobs in a separate `company_jobs` table.

Now, there is multiple problems with these

- Using Redis we already getting logs of all our jobs, with a much nicer UI and more information.
- The selection of the correct `companyId` is broken. In https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/Jobs/AbstractJob.php#L28 we are calling 

  ```php
  $databaseProxyManager = app()->make(DatabaseProxyManagerService::class);
  ```
  however, this is just

  ```php
          $user = $this->buildQuery()
              ->select(User::COL_COMPANY_ID)
              ->first();
  
          return $user->getCompanyId();
  ```
  see https://github.com/EnAccess/micropowermanager/blob/main/src/backend/app/Services/UserService.php#L132-L139 Which returns the first user's company id.

### Proposed Solution

The `AbstractJob` in it's current form doesn't provide any value. Either fix or remove it.

One idea would be to require all Jobs that inherit from `AbstractJob` to be provided their `companyId` from the caller. For example, when a Job is triggered from a Controller, the controller would then pass down the authenticated users company id. For scheduled jobs we would be looping over all company ids, creating on job per id.

Whether this is the best solution shall be investigated as part of the issue.

### Alternatives Considered

N/A

PR #864, opened 2025-08-21
Refactor jobs abstraction

beesaferoot:
Makes progress on #837

### Pull Request checklist


dmohns (commented):
Looks good. Only question: Can we entirely get rid of `CompanyJob` now? Not necessarily in this PR, as we might want to confirm everything is working as expected first. But we could create a follow up task.

beesaferoot:
> Looks good. Only question: Can we entirely get rid of `CompanyJob` now? Not necessarily in this PR, as we might want to confirm everything is working as expected first. But we could create a follow up task.

Yes, CompanyJob isn't used anywhere else. That would be the next step to fully closing #837

PR #873, opened 2025-08-25
feat:add export csv button for transaction and customer

munyanezaarmel:
added export csv button for transaction and customer

 closes:#84
image
image

### Pull Request checklist


beesaferoot:
Thanks for the PR @munyanezaarmel 

WDYT about using a less technical phrasing i.e "Export Customers", "Export Transactions"?

munyanezaarmel:
> Thanks for the PR @munyanezaarmel
> 
> WDYT about using a less technical phrasing i.e "Export Customers", "Export Transactions"?

this is good user experience what do you mean about this "we might also what to make it flexible for users to filter what to export" more context

munyanezaarmel:
@beesaferoot @dmohns  What are your thoughts? on these filters??
image
image

beesaferoot:
> > > > Thanks for the PR @munyanezaarmel
> > > > WDYT about using a less technical phrasing i.e "Export Customers", "Export Transactions"?
> > > 
> > > 
> > > this is good user experience what do you mean about this "we might also what to make it flexible for users to filter what to export" more context
> > 
> > 
> > The export currently assumes the user wants all the transactions and no filter.
> > Looking at the export on the backend, it supports filtering i.e `deviceType`, `provider`, `currency`, `format` etc
> > Not super important right now, maybe something we can consider.
> 
> So, if I heard you correctly, we don't need filters now.

What you've shown on the screenshots looks good to me. WDYT @dmohns?

dmohns:
> So, if I heard you correctly, we don't need filters now.

I'd the filters are nice-to-have but not mandatory. As @beesaferoot mentioned our main use-case is that customers can export all transactions and users (for example to load them into Excel/Gsheet/etc and do some data analytics). 

So, as long as the frontend filters are optional, i.e. users can leave them empty and just click export (and the UI is self-explanatory enough that this is the case) it be good.

munyanezaarmel:
@beesaferoot please check again

munyanezaarmel:
@dmohns , would you mind adding the other translations in the other language files? I changed en.json and it requires other translations to be added.

dmohns:
> @dmohns , would you mind adding the other translations in the other language files? I changed en.json and it requires other translations to be added.

You can just use Google Translate (or any other translator service) to get the other translation. I would just be doing the same 😂

dmohns (approved):
Nice, I like this 👍

beesaferoot (approved):

Issue #866, opened 2025-08-21
[Feature Request]: Apply `larastan` to entire PHP code


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Currently we are only running `larastan` on the `app` directory. Errors like: https://github.com/EnAccess/micropowermanager/issues/865 could easily be avoided if we ran on entire backend code.

The main task here, is to apply Larastan on the `tests` directory.

### Proposed Solution

Apply `larastan` on `src/backend`

### Alternatives Considered

N/A

PR #874, opened 2025-08-25
Apply Larastan to `packages` directory

dmohns:

Makes progress on : #866 

The packages directory has a billion Larastan errors. Hence we taking it slowly and try to Fix Larastan Level **1** and **2** here.

### Pull Request checklist


src/backend/packages/inensus/kelin-meter/src/Http/Resources/KelinMeterStatusResource.php
```diff
@@ -6,25 +6,26 @@
...
+                    'positiveReactiveDailyPower' => $this->dlZw, // @phpstan-ignore property.notFound
+                    'invertedActiveDailyPower' => $this->dlFy, // @phpstan-ignore property.notFound
+                    'invertedReactiveDailyPower' => $this->dlFw, // @phpstan-ignore property.notFound
+                    'energyRemain' => $this->energyRemain, // @phpstan-ignore property.notFound
+                    'moneyRemain' => $this->moneyRemain, // @phpstan-ignore property.notFound
+                    'meterStatus' => $this->meterStatus, // @phpstan-ignore property.notFound
+                    'openCoverCount' => $this->openCoverCount, // @phpstan-ignore property.notFound
+                    'openTerminalCount' => $this->openTerminalCount, // @phpstan-ignore property.notFound
+                    'owner' => $this->owner, // @phpstan-ignore property.notFound
+                    'meterAddress' => $this->meterAddress, // @phpstan-ignore property.notFound
```
alaincodes24, line 28: I wonder if there might be a way to have these properties recognized by PHPStan without relying on `@phpstan-ignore...`. One of the conversations we had, we discouraged the use of `@phpstan-ignore`

alaincodes24 (commented):
Thanks for this big refactoring!

alaincodes24 (approved):

dmohns:
@alaincodes24 I did some more changes. Essentially, I'm now allowing the Interfaces to be much more generic. The additional effort of making Morph Relations work with interfaces is just not worth it. It just makes our code messy with very little gain.

A small side learning of this exercise was that `ManufacturerTransactionInterface` and `PaymentProviderTransactionInterface` should probably **not** be interfaces 🙈 They both just define a handful of functions which are a 1-to-1 copy/paste in each of the implementing models. We can probably convert the abstraction to inheritance using something like `BasePaymentProviderTransaction` (or even composition using traits).

alaincodes24:
@dmohns Great insights! I completely agree. You're absolutely right about the interfaces becoming unnecessarily complex for limited benefit. When interfaces are just 1-to-1 copy/paste across implementations, they're not providing real value - just extra overhead. 

The inheritance/composition approach makes much more sense. `BasePaymentProviderTransaction` would eliminate repetitive method definitions. Trait-based approach like `HasPaymentProviderRelations` provides much cleaner relationship handling.

Should we create an issue for converting the abstraction to inheritance?

Issue #79, opened 2024-05-30
Filter customers by agent 


dmohns:
this will allow us to filter which customers ‘belong’ to which agents so we can calculate their collection commission

munyanezaarmel:
@dmohns @beesaferoot Let's split this to make it easy and quick. I have a question: how are we going to link customer and agent, i.e., when creating an agent, be able to add their customers, or when creating a customer, be able to add the agent they belong to?

beesaferoot:
> [@dmohns](https://github.com/dmohns) [@beesaferoot](https://github.com/beesaferoot) Let's split this to make it easy and quick. I have a question: how are we going to link customer and agent, i.e., when creating an agent, be able to add their customers, or when creating a customer, be able to add the agent they belong to?

The `agents` table (via Agent model) has a relationship with a customer using the `people' table (via Person model)


You can use this to filter customers by agent. 


<img width="1071" alt="Image" src="https://github.com/user-attachments/assets/2863244a-98f0-4fd8-b0dd-6dd9534e6ba9" />

munyanezaarmel:
If I understand you correctly, "person" refers to the customer. In that case, how does this relate to the data we are retrieving for the customer list? We intend to retrieve data for the customer list, correct?
`
    "data": [
        {
            "id": 1,
            "person_id": 251,
            "mini_grid_id": 1,
            "agent_commission_id": 1,
            "name": "Russel",
            "email": "irma.cummerata@example.com",
            "fire_base_token": "-",
            "balance": 900041.87,
            "commission_revenue": 0,
            "due_to_energy_supplier": 0,
            "created_at": "2025-03-06T09:07:13.000000Z",
            "updated_at": "2025-03-06T09:08:19.000000Z",
            "connection": "tenant",
            "person": {
                "id": 251,
                "title": "Dr.",
                "education": "MicroPowerManager Agent",
                "name": "Russel",
                "surname": "Fay (Agent - Mafia Village)",
                "birth_date": "1990-10-04",
                "sex": "0",
                "nationality": null,
                "is_customer": 0,
                "deleted_at": null,
                "created_at": "2025-03-06T09:07:13.000000Z",
                "updated_at": "2025-03-06T09:07:13.000000Z",
                "additional_json": null,
                "addresses": [
                    {
                        "id": 501,
                        "email": "rath.rosetta@example.net",
                        "phone": "+12816549110",
                        "street": "57 Hamill Street",
                        "city_id": 1,
[...]

PR #879, opened 2025-08-28
feat:filter customer by agent

munyanezaarmel:
filter customer by agent
closes:#79

image

### Pull Request checklist


beesaferoot (commented):
Looking good, and the filtering works. Nice! 

One comment I have is on the pagination ui when you filter, it seem to still be the one from the initial customer list. 


Screenshot 2025-09-01 at 12 40 40 PM

dmohns:
Cool, let's go with Option 1 then 👍 

What I mean is: When a User selects a certain Agent from the filter, we will show a filter list. What happens if they send click "Export Customers" will this export all customer or only the customers for the filtered agent?

(Sorry, I didn't have the time run your PR locally, to find it out myself 🙈 )

munyanezaarmel:
> Cool, let's go with Option 1 then 👍
> 
> What I mean is: When a User selects a certain Agent from the filter, we will show a filter list. What happens if they send click "Export Customers" will this export all customer or only the customers for the filtered agent?
> 
> (Sorry, I didn't have the time run your PR locally, to find it out myself 🙈 )

It will export all customers. Do you think we should change this?

dmohns:
> > Cool, let's go with Option 1 then 👍
> > What I mean is: When a User selects a certain Agent from the filter, we will show a filter list. What happens if they send click "Export Customers" will this export all customer or only the customers for the filtered agent?
> > (Sorry, I didn't have the time run your PR locally, to find it out myself 🙈 )
> 
> It will export all customers. Do you think we should change this?

It's more of a UX comment:

As a user, this is what I would expect to happen. However, I assume this isn't as straightforward (because we would need to change the underlying backend export API, right @beesaferoot ) Maybe as a quick "workaround" to make it more clear, we can change the name of the Export button to: "Export All Customers"? We could wait and see if any users requests this as a feature to export list of customers filtered by Agent.

dmohns (approved):
Thanks!

PR #890, opened 2025-09-01
Update dependencies 20250901

dmohns:

With updated version of PHPStan and Laravel, we no longer need redundant `@var` type hints.

For example, in 

```
/** @var Meter|null $result */
        $result = $this->newQuery()->where('serial_number', '=', $meterSerialNumber)->first();
```

the type hint `@var Meter|null $result` can be omitted. Larastan now understands it, when using `newQuery()`.

Additionally, changing the way `resolve()` is used in the app. 

- Use `resolve(...::class)` in the code where possible, to get much better type hinting.
- Register `aliases` in the service container. This can be used, when an API is resolved via Database entries (where we shouldn't use `class-string`, as these might change)

### Pull Request checklist

### DIFF
--- a/src/backend/packages/inensus/angaza-shs/src/Providers/AngazaSHSServiceProvider.php
+++ b/src/backend/packages/inensus/angaza-shs/src/Providers/AngazaSHSServiceProvider.php
@@ -25,7 +25,8 @@ class AngazaSHSServiceProvider extends ServiceProvider {
         $this->mergeConfigFrom(__DIR__.'/../../config/angaza-shs.php', 'angaza-shs');
         $this->app->register(EventServiceProvider::class);
         $this->app->register(ObserverServiceProvider::class);
-        $this->app->bind('AngazaSHSApi', AngazaSHSApi::class);
+        $this->app->bind(AngazaSHSApi::class);
+        $this->app->alias(AngazaSHSApi::class, 'AngazaSHSApi');
     }
 
     public function publishConfigFiles(): void {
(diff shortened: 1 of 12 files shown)


src/backend/app/Exceptions/Export/CsvNotSavedException.php
```diff
@@ -3,6 +3,8 @@
 namespace App\Exceptions\Export;
 
 class CsvNotSavedException extends \Exception {
-    /** @var string */
+    /**
+     * @var string
+     */
```
beesaferoot, line 8: Did you mean to remove this instead

src/backend/app/Models/ProtectedPage.php
```diff
@@ -5,6 +5,8 @@
 use App\Models\Base\BaseModelCentral;
 
 class ProtectedPage extends BaseModelCentral {
-    /** @var string */
+    /**
```
beesaferoot, line 8: Same here

beesaferoot (dismissed):

src/backend/packages/inensus/africas-talking/src/Providers/AfricasTalkingServiceProvider.php
```diff
@@ -24,7 +24,8 @@ public function register(): void {
         $this->mergeConfigFrom(__DIR__.'/../../config/africas-talking.php', 'africas-talking');
         $this->app->register(EventServiceProvider::class);
         $this->app->register(ObserverServiceProvider::class);
-        $this->app->bind('AfricasTalkingGateway', AfricasTalkingGateway::class);
+        $this->app->bind(AfricasTalkingGateway::class);
+        $this->app->alias(AfricasTalkingGateway::class, 'AfricasTalkingGateway');
```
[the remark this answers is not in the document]
dmohns: Haha, yeah, from this code alone it's not really clear. The idea here is to be able to call

```
resolve(AfricasTalkingGateway::class)
```

This seems to be the "Laravel-way" of doing it, like here: https://laravel.com/docs/11.x/container#the-make-method
(`resolve` is just an alias to `$this->app->make(...)`). This is nice, because we get auto-completion etc... as it understands that the return will be `AfricasTalkingGateway`.

I'm adding the alias as a precaution, in case there is a place left where we still call`resolve('AfricasTalkingGateway`)`.


**Note:** There is a difference for manufacturer API's. These are read from the database, i.e. the `api_name` column on `Device`. Here, we don't want to write the full Laravel class path into `api_name` column. But rather an alias. (similar to how we do in morph maps)

beesaferoot (approved):

PR #901, opened 2025-09-05
Update Larastan to version 3

dmohns:

Updates to Larastan version 3 (and simultaneously PHPStan 2). Newer version reports some additional errors which are also getting fixed in this PR.

### Pull Request checklist

### DIFF
--- a/src/backend/app/Helpers/MailHelperMock.php
+++ b/src/backend/app/Helpers/MailHelperMock.php
@@ -2,9 +2,6 @@
 
 namespace App\Helpers;
 
-use App\Exceptions\MailNotSentException;
-use PHPMailer\PHPMailer\Exception as PHPMailerException;
-
 class MailHelperMock implements MailHelperInterface {
     /**
      * @param string      $to
--- a/src/backend/app/Misc/LoanDataContainer.php
+++ b/src/backend/app/Misc/LoanDataContainer.php
@@ -5,7 +5,6 @@ namespace App\Misc;
 use App\Events\PaymentSuccessEvent;
 use App\Exceptions\Meters\MeterIsNotAssignedToCustomer;
 use App\Exceptions\Meters\MeterIsNotInUse;
-use App\Exceptions\Meters\MeterNotFound;
 use App\Models\AssetPerson;
 use App\Models\AssetRate;
 use App\Models\Meter\Meter;


src/backend/app/Http/Controllers/Reports.php
```diff
@@ -444,7 +437,7 @@ private function addConnectionGroupsToXLS(
 
             $meters = $connectionGroup->meters()->get();
 
-            if ($meters !== null) {
+            if (!$meters->isEmpty()) {
```
beesaferoot, line 440: Do we need to have a null check just to be safe here? i.e

```php
if(isset($meters) and !$meters->isEmpty()) {
}
```

src/backend/app/Http/Controllers/RevenueController.php
```diff
@@ -152,7 +152,7 @@ public function revenueData(Request $request): ApiResource {
                     'average_revenue_per_month' => '-',
                 ];
             }
-        } elseif ($targets !== null && $targetType === 'mini-grid' && isset($targets->subTargets)) {
+        } elseif ($targetType === 'mini-grid' && isset($targets->subTargets)) {
```
beesaferoot, line 155: I'm concerned about removing the null check, is it guaranteed not never ever be null?

src/backend/app/Http/Controllers/Reports.php
dmohns: But, yeah, it's technically changing the logic. If, somehow `$meters` IS null, then modified code will fail.

Let me add your suggestions 👍

dmohns: Just tried that, but

```
Variable $meters in isset() always exists and is not nullable. 
```

🙈 

The reason here is: Earlier in the function we have:

```php
if (!isset($connectionGroup->name)) {
    continue;
}
```

The only way `$meter` could ever be `null` is when `$connectionGroup` is unset. Which is guarded by the above.

src/backend/app/Http/Controllers/RevenueController.php
beesaferoot: I missed that 😄. Thanks. I can be less woried now

beesaferoot (approved):

Issue #904, opened 2025-09-08
[Feature Request]: Improve the Customer Export feature


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

The recently added Export Customer feature is great, but has some hiccups.

1. The export dialogs prompts the user to Provider Currency and Timezone. These don't make any sense for customer export and cause confusion.
2. It should be possible to filter the customers by geographical region (MiniGrid or Village)

Additional feedback provided by users

- Both exports don’t seem to have our currency nor allow us to choose the option **All** for certain fields that have the option (i.e, device type, provider, status);
- The field Serial Number on the transactions export doesn’t seem to be necessary because we want the entire list of transactions for all serial numbers involved in that period (perhaps the field can stay as optional);
- When we choose “Excel” on Format it still downloads as a CSV file;
- The drop down filters should only show selectables with at least one entry. (For example, a tenant not using WaveCom Transactions, shouldn't see this in the dropdown)

Additionally, users encountered the following error

<img width="3659" height="1869" alt="Image" src="https://github.com/user-attachments/assets/9dda2d83-811d-48d7-89c8-d97f0ce133fe" />

### Proposed Solution

Please address the above mentioned improvements.

### Alternatives Considered

N/A

beesaferoot:
@munyanezaarmel  it would be great if you could address feedback on the frontend and then create an issue specifying the behaviour the frontend is expecting that is not currently correct on the backend.

beesaferoot:
> > [@munyanezaarmel](https://github.com/munyanezaarmel) it would be great if you could address feedback on the frontend and then create an issue specifying the behaviour the frontend is expecting that is not currently correct on the backend.
> 
> Yes, I’m fixing them. I just wanted some advice on whether we need the filters or if we can ignore them.

Okay, I understand your comment now. 

I think we still need them judging from the user feedback we just need to polish the behaviour a bit. 

[...]

PR #919, opened 2025-09-17
feat:Improve the Customer Export feature #904

munyanezaarmel:
Improve the Customer Export feature 
closes:#904
image
image

### Pull Request checklist

### DIFF
--- a/src/frontend/src/modules/Transactions/FilterTransaction.vue
+++ b/src/frontend/src/modules/Transactions/FilterTransaction.vue
@@ -23,19 +23,6 @@
             </md-field>
           </div>
           <div
-            class="md-layout-item md-xlarge-size-100 md-large-size-100 md-medium-size-100 md-small-size-100 md-xsmall-size-100"
-          >
-            <md-field>
-              <label for="serial_number">Serial Number</label>
-              <md-input
-                type="text"
-                name="serial_number"
-                v-model="filter.serial_number"
-                v-on:keyup.enter="submitFilter"
-              ></md-input>
-            </md-field>
-          </div>
-          <div
             v-if="selectedDevice === 'meter'"
             class="md-layout-item md-xlarge-size-100 md-large-size-100 md-medium-size-100 md-small-size-100 md-xsmall-size-100"
           >
@@ -158,6 +145,9 @@ import moment from "moment-timezone"
 import store from "@/store/store"
 import { TransactionExportService } from "@/services/TransactionExportService"
 import { notify } from "@/mixins"
+import { MiniGridService } from "@/services/MiniGridService"
+import { CityService } from "@/services/CityService"
+import { CurrencyListService } from "@/services/CurrencyListService"
 
 export default {
   name: "FilterTransaction",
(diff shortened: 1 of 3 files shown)


src/frontend/src/modules/Transactions/FilterTransaction.vue
```diff
@@ -322,6 +340,71 @@ export default {
...
+        }
+
+        const response = await this.customerExportService.exportCustomers(data)
+        const blob = new Blob([response.data])
+        const downloadUrl = window.URL.createObjectURL(blob)
+        const a = document.createElement("a")
+        a.href = downloadUrl
+        const contentDisposition = response.headers["content-disposition"]
+        const fileNameMatch = contentDisposition?.match(/filename="(.+)"/)
+        a.download = fileNameMatch ? fileNameMatch[1] : "export_customers.csv"
```
beesaferoot, line 397: You are defaulting to only csv here I see. Don't we need to handle the excel export type as well?

src/frontend/src/modules/Transactions/FilterTransaction.vue
```diff
@@ -322,6 +340,71 @@ export default {
...
+          data.isActive = this.exportFilters.isActive
+        }
+        if (this.exportFilters.miniGrid) {
+          data.miniGrid = this.exportFilters.miniGrid
+        }
+        if (this.exportFilters.village) {
+          data.village = this.exportFilters.village
+        }
+        if (this.exportFilters.deviceType) {
+          data.deviceType = this.exportFilters.deviceType
```
beesaferoot, line 384: How we using this anywhere? Doesn't seem to be reference anywhere.

munyanezaarmel: > Also, I need to know what value "miniGrid" and "village" mean here? Are they ids?

Yes, miniGrid` and `village` should be IDs, I guess

beesaferoot:
@munyanezaarmel The updates looks good, can you also add that to the transactions export?

munyanezaarmel:
> @munyanezaarmel The updates looks good, can you also add that to the transactions export?

Oh sorry, fixed, let's check again

beesaferoot:
> Two more comments on Transaction export
> 
> image
> "Currency" and "Time Zone" are weird here. They are not filters (!), but "settings" for the export. But in the backend, we are not going to do Currency conversion for the export anyways, so I think we can just remove it.
> 
> Similar thoughts for timezone... Even tho it could be nice for a user to configure which timezone the export is in, this adds to much complexity. I'd say we just export ISO8601 timestamps ( for example `2025-09-22T08:53:41Z`) which include the timezone.
> 
> So, as far as this PR is concerned, you can just remove the two. The rest is going to be done in the backend parity PR. @beesaferoot let me know what you think

The `timezone` filter is tricky as you rightly pointed out, however, the `currency` filter can mean two things. Since users can update their currency settings and make transactions with that it means you can have a transaction list with multiple currencies.Filtering that automatically on the backend might surprise users.

munyanezaarmel:
@beesaferoot @dmohns let's check again

beesaferoot (approved):
I

beesaferoot:
> Cool, the UI looks great now 🎉 The rest of the changes will now be backend ones, right?

Yes @dmohns I will be handling that part.

PR #999, opened 2025-10-17
Define API level RBAC layer and policies

### LINKED ISSUE #998: [Feature Request]: RBAC support on the API level
### Preflight Checklist

### Problem Description

There have been quite a number of requests to support role-based access control on MPM to further extend the usuability of the platform as a CRM tool. 

### Proposed Solution

The goal for this issue is to define and implement role based access control layer over the APIs. This will build the ground work needed to fully support RBAC on the platform. 

The design here is to have users with assign roles and calls to endpoints correctly authorize or reject requests with role based privileges. 

### Alternatives Considered

N/A

### Additional Information

_No response_

beesaferoot:
Closes #998 

### Pull Request checklist

### DIFF
--- a/src/backend/app/Http/Controllers/UserController.php
+++ b/src/backend/app/Http/Controllers/UserController.php
@@ -52,6 +52,14 @@ class UserController extends Controller {
         // Handle role assignment if provided
         if ($request->has('roles')) {
             $roles = (array) $request->input('roles', []);
+
+            // Validate each role assignment permission
+            foreach ($roles as $roleName) {
+                if (!$request->user()->can('assignRole', [User::class, $roleName])) {
+                    abort(403, "You don't have permission to assign the role: {$roleName}");
+                }
+            }
+
             // Ensure user has at least one role
             if ($roles !== []) {
                 $user->syncRoles($roles);
--- a/src/backend/app/Policies/UserPolicy.php
+++ b/src/backend/app/Policies/UserPolicy.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace App\Policies;
+
+use App\Models\User;
+
+class UserPolicy {
+    public function view(User $currentUser, User $targetUser): bool {
+        return $currentUser->can('users');
+    }
+
+    public function create(User $currentUser): bool {
+        return $currentUser->can('users');
+    }
+
+    public function update(User $currentUser, User $targetUser): bool {
+        // Users can't update themselves to prevent privilege escalation
+        if ($currentUser->id === $targetUser->id) {
+            return false;
+        }
+
+        // Check if target user is an owner
+        if ($targetUser->hasRole('owner')) {
+            // Only users with 'users.manage-owner' permission can update owners
+            return $currentUser->can('users.manage-owner');
+        }
+
+        // Check if target user is an admin
+        if ($targetUser->hasRole('admin')) {
+            // Only users with 'users.manage-admin' permission can update admins
+            return $currentUser->can('users.manage-admin');
+        }
+
+        // For other roles (financial-manager, user), basic 'users' permission is enough
+        return $currentUser->can('users');
+    }
+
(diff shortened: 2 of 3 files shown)


dmohns (commented):
Hey @beesaferoot 

trying to make sense out of what is happening here. It looks great, but I'm not an expert with Spatie Permissions, so bear with me 😅 

Dropping a few questions/observations

- Can you explain a bit more what the approach here is? (I understand that this may change in the future, but at least for the current) From what I understand the Permissions are essentially hardcoded in the MPM Code base. What about roles? I understand admin's can assign different roles to users, but can they also change the roles?
- There seems to be something wrong with the migrate for existing users. For existing users, it assigns the Owner permission to the first user and admin to all others. However, these roles are never assigned any permissions.
- I think the above also applies for newly created companies. The roles will be assigned but no permissions seems to be added.

For example, when I checkout this branch and run with an existing Demo setup (not a newly seeded)

```json
{
  "access_token": "...",
  "token_type": "bearer",
  "expires_in": 3600,
  "user": {
    "id": 1,
    "name": "Demo Company Admin",
    "company_id": 1,
    "email": "demo_company_admin@example.com",
    "created_at": "2025-09-23T09:41:28.000000Z",
    "updated_at": "2025-09-23T09:41:28.000000Z",
    "roles": [
      {
        "id": 1,
        "name": "owner",
        "guard_name": "api",
        "created_at": "2025-10-17T13:37:11.000000Z",
        "updated_at": "2025-10-17T13:37:11.000000Z",
        "pivot": {
          "model_type": "users",
          "model_id": 1,
          "role_id": 1
        },
        "permissions": []
      }
    ],
    "permissions": []
  },
  "roles": [
    "owner"
  ],
  "permissions": []
}
```

dmohns:
There seems to be some confusion with existing Tables/models.

Apparently, we already have `Role/Roles` and `Role/RoleDefinition`. As a result, the spatie roles go into a table called, `roles_permission`, as per `config`. This is a bit confusing when ready `spatie-permission` docs 😅 

Could we also clean up (delete `Role/Roles` and `Role/RoleDefinition`) and use the default `roles` table for spatie roles?

dmohns:
> There are two important parts:
> 
> 1. Roles
>    These are tied to user/agent accounts. Right now we have to seed them manually via `RoleBasePersimmisionSeeder.php` but the goal is to manage them dynamically from the UI.
> 2. Permissions
>    These are defined as policies and are tied to specific route guards. I don't think they is a way create them dynamically, they have to be created in the codebase to match the permission layers that best reflect the platform.
> 
> The nice thing about these policies are they can added during route definition i.e
> 
> ```
> Route::get('/{personId}/meters/geo', [MeterGeographicalInformationController::class, 'show'])->middleware('permission:customers.view');
> ```

Understood 👍 

> The issue here, I believe, is because I went with a manual way of seeding the roles and permissions beforehand. You need to have run the command `php artisan db:seed --class=Database\\Seeders\\RoleBasePermissionSeedeer --database=tenant` before running the migration.
> 
> > I think the above also applies for newly created companies. The roles will be assigned but no permissions seems to be added.
> 
> This should be resolved by following the steps above.

Got it. But then my comment here would be the following: I'm not sure if we need a fully customisable (by MPM users) Role and Permission Model for MPM. I think it will get _really_ complicated if want to give users a full flexibility.

WDYT about going with a set of predefined roles that users can choose from? I.e. something like `owner`, `admin`, `agent` (like we already have), `editor` and `reader` ?

dmohns (commented):
With the new simplified role model of Owner, Admin, Financial Manager and User we don't need to break down permissions into CRUD operations anymore.

For example, a single `payments` permission would suffice. We can use fine-grained roles for edge cases, like users management w.r.t. admin and owner.

dmohns (commented):
few comments

Issue #992, opened 2025-10-16
[Feature Request]: Deprecate `BaseModel` from packages


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Certain packages like `ticket`, `spark-meter`, `steama-meter` and potentially others which with their own copy of `BaseModel.php`.

This is redundant and can cause confusion.

### Proposed Solution

Please migrate all packages to use the shared `BaseModel` from the root.

### Alternatives Considered

N/A

F4tal1t:
i will try to fix this one @dmohns

PR #1007, opened 2025-10-19
Deprecate package BaseModels, migrate to shared root BaseModel, remove redundancies

F4tal1t:
Closes: #992 

This PR deprecates redundant `BaseModel.php` files in packages like `ticket`, `spark-meter`, `steama-meter`, `bulk-registration`, and `mesomb-payment-provider`. All affected models now extend the shared `BaseModel` from the root `BaseModel.php`, eliminating duplication. The root `BaseModel` was updated to include the `resolveChildRouteBinding` method for compatibility. This centralizes the base model, reducing maintenance overhead and potential inconsistencies across the codebase.

None as of now.

Testing was conducted in the dev container (Ubuntu 24.04.2 LTS). Syntax checks (php -l) passed on the root `BaseModel` and sample migrated models (e.g., `Ticket.php`). File integrity was verified: redundant `BaseModel.php` files were deleted, and all extends were updated. Prettier ran successfully on frontend code. Full composer install failed due to PHP version mismatch (8.0 vs. required 8.2+), but targeted validation confirmed no breaking changes. Recommend running php artisan test in a compatible environment for end-to-end validation

### Pull Request checklist

### DIFF
--- a/src/backend/packages/inensus/bulk-registration/src/Models/CsvData.php
+++ b/src/backend/packages/inensus/bulk-registration/src/Models/CsvData.php
@@ -4,6 +4,6 @@ namespace Inensus\BulkRegistration\Models;
 
 use App\Models\Base\BaseModel;
 
-class CsvData extends \App\Models\Base\BaseModel {
+class CsvData extends BaseModel {
     protected $table = 'bulk_registration_csv_datas';
 }
--- a/src/backend/packages/inensus/spark-meter/src/Models/SmCredential.php
+++ b/src/backend/packages/inensus/spark-meter/src/Models/SmCredential.php
@@ -2,6 +2,8 @@
 
 namespace Inensus\SparkMeter\Models;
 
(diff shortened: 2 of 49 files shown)


src/backend/app/Models/Base/BaseModel.php
```diff
@@ -25,4 +25,9 @@ public function __construct(array $attributes = []) {
 
         parent::__construct($attributes);
     }
+
+    public function resolveChildRouteBinding($childType, $value, $field) {
+        // TODO: Implement resolveChildRouteBinding() method.
+        throw new \Exception('Method resolveChildRouteBinding() not yet implemented.');
+    }
```
dmohns: It looks like this isn't used at all, or am I missing something? Probably it's save to remove it.

F4tal1t: saw it being used in one of the BaseModel files i deleted , thought it was needed , will remove it now as  you said.

PR #1009, opened 2025-10-20
Improve development setup

dmohns:

Here are a set of improvements to our local development setup. 

There are a lot of tool PHP tools which can help us develop MicroPowerManager which we unfortunately cannot use right now.
A great example of such tool is the famous [Laravel IDE Helper](https://github.com/barryvdh/laravel-ide-helper).

Due to the way our Multi-Tenancy is implemented a lot of helper function cannot be called directly, but always need to be wrapped in a `databaseProxyManager->runForCompany(...)`.

For example, when running `php artisan model:show "Person\Person"` yields

```
   ErrorException 

  Undefined array key "driver"

  at vendor/laravel/framework/src/Illuminate/Database/DatabaseManager.php:203
```

To circumvent this, we can just set the default value for `tenant` connection to the one of the DemoCompany.

However, we should only ever do this in development environments.

Additionally, some improvement to our Code Style rules:

- `phpdoc_no_alias_tag`: Even tho `property-read` and `property` are aliases, using `property-read` provides some value when it's used for magic properties.
- `phpdoc_align` adapted to work with `property-read`

### Pull Request checklist

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -36,6 +36,10 @@ return (new PhpCsFixer\Config())
             'on_multiline' => 'ensure_fully_multiline',
         ],
         'single_line_empty_body' => true,
+        'phpdoc_no_alias_tag' => false,
+        'phpdoc_align' => [
+            'tags' => ['method', 'param', 'property', 'property-read', 'return', 'throws', 'type', 'var'],
+        ],
         // disabled for editor integration
         'no_unused_imports' => false,
     ])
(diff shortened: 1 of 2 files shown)


beesaferoot:
> > Still a little unsure what the purpose for the tool `direnv` would be for. Is this to simply export the database env variables?
> 
> Yeah, exactly. This makes it, so you can run PHP Tooling outside of the Docker container!

😄 Still confused about how this works. Just install the tool, and you get the property out of the box? Because I don't see any setup done on our end besides the `.envrc` file

beesaferoot (approved):
Thanks for the clarification @dmohns  

This is great dx improvement

PR #1024, opened 2025-10-22
Create PROXY ONLY subnet once per project

dmohns:

PROXY ONLY subnet should be unique per project. Updating our Terraform module to account for that, in scenarios where multiple environments are managed in the same GCP project.

To achieve this, we are introducing a new variable `network_proxy_only_subnet_cidr_range`. When the proxy only subnet is managed (created) this CIDR range will be used, else is assumed to be existing.

### Pull Request checklist

### DIFF
--- a/terraform/gcp_kubernetes/main.tf
+++ b/terraform/gcp_kubernetes/main.tf
@@ -135,8 +135,8 @@ resource "google_compute_firewall" "allow_proxy_connection" {
 }
 
 moved {
-    from = google_compute_firewall.rules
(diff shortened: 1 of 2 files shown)


dmohns:
> Of course I barely understand what these are for 😄

To be very honest. I don't understand it either. I'm just rigorously following: https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balance-ingress#prepare-environment

They say we need a proxy only subnet to deploy internal load balancing. So I'm creating it 😂

Issue #1052, opened 2025-10-31
[Feature Request]: Simplify plugin and package creation


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Currently, the initial package creation is tedious and confusing for new users.

Few pain points

- The package generation is based on an outdated starter, which we don't have control over https://github.com/inensus/Package-Development-Starter-Pack
- After initial command, developers have to manually copy files around, which is error prone
- The involved shell script does not work on Windows machines (if called natively, not inside Docker)
- Confusions are around migration vs stub migration files

### Proposed Solution

- Decomission the dependency on an external Github repo, and manage the template inside MPM repository
- Convert the Shell script to a full PHP native artisan command

### Alternatives Considered

N/A

PR #1075, opened 2025-11-06
Manage Plugin Starter inside MPM repository

dmohns:

Makes progress on: #1052 

In this PR we are doing two main things

- Move the `package-starter.sh` shell script into native PHP code
- Track the content of package starter inside the MPM repo instead of an external one

There is some subtle, but impactful side effect. Because we are no longer `git clone`-ing the starter, but copying it from a folder outside of `backend`, the plugin starter can no longer be run inside the MPM Backend Docker container.

This might seem like a drawback at first glance. But it's actually an improvement. The fact that the plugin starter is run from inside the MPM Docker container prevents any changes to files outside of `backend`.
As a result, Users have to perform a long list of manual steps that cannot be automated.

After the current PR, users _still_ have to do many copy/paste steps. But it will allow us to greatly improve this flow in the future.

### Pull Request checklist


beesaferoot (approved):
Game changer 🪄  Thanks @dmohns 


Minor comment: 
Is there a way to make it clear the command `php artisan micropowermanager:new-package {plugin-name}` should be ran inside the `src/backend` to work?

beesaferoot (approved):

PR #1147, opened 2025-11-27
handle rbac edge case + improve overview page

beesaferoot:
Follow up to #1140 

Screenshot 2025-11-27 at 6 27 07 PM

### Pull Request checklist

### DIFF
--- a/src/backend/app/Http/Resources/UserListResource.php
+++ b/src/backend/app/Http/Resources/UserListResource.php
@@ -19,7 +19,7 @@ class UserListResource extends JsonResource {
             'name' => $this->name,
             'email' => $this->email,
             'roles' => $this->roles->pluck('name'),
-            'address_details' => $this->addressDetails->toArray(),
+            'address_details' => $this->addressDetails?->toArray() ?? [],
         ];
     }
 }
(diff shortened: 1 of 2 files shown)


src/backend/app/Policies/UserPolicy.php
```diff
@@ -18,15 +18,16 @@ public function create(User $currentUser): bool {
...
         // Check if target user is an owner
         if ($targetUser->hasRole('owner')) {
             // Only users with 'users.manage-owner' permission can update owners
+            // ensure at least one owner exists
+            if (User::whereHas('roles', function ($query) {
+                $query->where('name', 'owner');
+            })->count() === 1) {
+                return false;
+            }
+
```
dmohns: Somehow Github swallowed my comment... Here it is again:

Currently, Owner users cannot modified their own roles at all when they are the only Owner user.

This scenario might occur, when an owner wants to give themselves the admin role in addition to owner to prepare the transfer of owner to someone else. In the current way, they would have to first make someone else an owner, and then remove their own owner role at the same time as giving another role

beesaferoot: But why not just create an Admin account login and set it up directly before giving them access? Why do they need to "be" and admin first?

Issue #1119, opened 2025-11-18
[Feature Request]: Customer transaction list should show "recent" customers first in default sorting


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

When a new customer is created, in the current default view, that customer is displayed at the end of the list. This is confusing. For example, in transaction views most recent transactions are shown first (by default.

### Proposed Solution

Please reverse the default sorting of customer list (based on recency). Please note: https://github.com/EnAccess/micropowermanager/issues/1117 that customers should be able to change the sorting as required.

### Alternatives Considered

N/A

munyanezaarmel:
Hy @dmohns 

Quick update on the sorting issues (#1117, #1118, #1119):

All three require **backend API extensions** to work properly. The API endpoints need to accept `sortBy` and `sortOrder` parameters.
Let me know and I can proceed with the implementation.

<img width="1072" height="178" alt="Image" src="https://github.com/user-attachments/assets/715bcdde-d967-4d09-b357-aad5520cf56f" />

dmohns:
> All three require **backend API extensions** to work properly. The API endpoints need to accept `sortBy` and `sortOrder` parameters. Let me know and I can proceed with the implementation.

Alright, this makes sense. I assume this will become even more relevant, as we apply the possibility to order to other lists as well.

@beesaferoot @munyanezaarmel do you know if there is a consistent way to add sorting capabilities to all list endpoints? Without it being a massive manual effort?

beesaferoot:
> > > All three require **backend API extensions** to work properly. The API endpoints need to accept `sortBy` and `sortOrder` parameters. Let me know and I can proceed with the implementation.
> > 
> > 
> > Alright, this makes sense. I assume this will become even more relevant, as we apply the possibility to order to other lists as well.
[...]

PR #1155, opened 2025-12-02
Add sort functionality to Customer/Transaction endpoints

beesaferoot:
Makes progress on #1119 

### Pull Request checklist

### DIFF
--- a/src/backend/app/Models/Person/Person.php
+++ b/src/backend/app/Models/Person/Person.php
@@ -62,7 +62,7 @@ use Illuminate\Support\Facades\DB;
  * @property-read Collection<int, Roles>           $roleOwner
  * @property-read Collection<int, Ticket>          $tickets
  */
-class Person extends BaseModel implements \Stringable, HasAddressesInterface, RoleInterface {
+class Person extends BaseModel implements HasAddressesInterface, RoleInterface {
     /** @use HasFactory<PersonFactory> */
     use HasFactory;
     use SoftDeletes;


dmohns (commented):
Not sure if I like this solution 🙈 

Couple of concerns I have:

- I seems like a lot of effort would be required to add sorting to existing queries. Need to maintain massive `switch` statements long term.
- Using `DB:raw(...)` should really be avoided whenever possible.

I think moving forward I would also like to add sorting to many more endpoints. Should we maybe review the decision to not consider https://spatie.be/docs/laravel-query-builder/v6/introduction ?

beesaferoot:
> Not sure if I like this solution 🙈
> 
> Couple of concerns I have:
> 
> * I seems like a lot of effort would be required to add sorting to existing queries. Need to maintain massive `switch` statements long term.
> * Using `DB:raw(...)` should really be avoided whenever possible.
> 
> I think moving forward I would also like to add sorting to many more endpoints. Should we maybe review the decision to not consider https://spatie.be/docs/laravel-query-builder/v6/introduction ?


I did explore introducing `laravel-query-builder` and wasn't convinced it would be any easier. Most of their example use the model directly; we have some complex relationship resolution that would still require the effort to adapt. 


Regarding the `switch` statements and `DB:raw`, these were unavoidable for the sort we wanted to achieve. If we rule out sort using data from relation models then we don't need any more than what was used in the `Transactions` Model.

src/backend/app/Services/PersonService.php
```diff
@@ -198,14 +205,61 @@ public function getAll(?int $limit = null, ?int $customerType = 1, ?int $agentId
...
+            ->defaultSort('-created_at')
+            ->paginate($limit);
     }
 
     /**
      * @return Collection<int, Person>|array<int, Person>
      */
     public function getAllForExport(?int $miniGrid = null, ?int $village = null, ?string $deviceType = null, ?bool $isActive = null): Collection|array {
-        $query = $this->person->newQuery()->with([
+        $query = Person::query()->with([
```
dmohns: as above

dmohns (approved):
Noice :shipit:

Issue #1157, opened 2025-12-02
[Bug]: Outstanding debts email failing due to error in Mailer


beesaferoot:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Expected Behavior

Emails with attachment should send without errors;

### Actual Behavior

The below error occurred when attempt to send outstanding debts email on Demo. 

```

ErrorException: Undefined array key 1 in /var/www/html/vendor/symfony/mime/Part/DataPart.php:39
--
Stack trace:
#0 /var/www/html/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(258): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(2, 'Undefined array...', '/var/www/html/v...', 39)
#1 /var/www/html/vendor/symfony/mime/Part/DataPart.php(39): Illuminate\Foundation\Bootstrap\HandleExceptions->Illuminate\Foundation\Bootstrap\{closure}(2, 'Undefined array...', '/var/www/html/v...', 39)
#2 /var/www/html/vendor/symfony/mime/Email.php(334): Symfony\Component\Mime\Part\DataPart->__construct(NULL, 'OutstandingDebt...', '')
#3 /var/www/html/vendor/laravel/framework/src/Illuminate/Mail/Message.php(327): Symfony\Component\Mime\Email->attach(NULL, 'OutstandingDebt...', '')
#4 /var/www/html/vendor/laravel/framework/src/Illuminate/Mail/Mailable.php(514): Illuminate\Mail\Message->attachData(NULL, 'OutstandingDebt...', Array)
#5 /var/www/html/vendor/laravel/framework/src/Illuminate/Mail/Mailable.php(496): Illuminate\Mail\Mailable->buildDiskAttachments(Object(Illuminate\Mail\Message))
#6 /var/www/html/vendor/laravel/framework/src/Illuminate/Mail/Mailable.php(214): Illuminate\Mail\Mailable->buildAttachments(Object(Illuminate\Mail\Message))
[...]

PR #1205, opened 2025-12-15
fix emaill attachment issue using attachFromStorageDisk

beesaferoot:
Closes: #1157 

### Pull Request checklist

### DIFF
--- a/src/backend/app/Services/AbstractExportService.php
+++ b/src/backend/app/Services/AbstractExportService.php
@@ -9,7 +9,6 @@ use App\Exceptions\Export\SpreadSheetNotSavedException;
 use App\Models\DatabaseProxy;
 use App\Models\User;
 use Carbon\Carbon;
-use Illuminate\Contracts\Filesystem\Filesystem;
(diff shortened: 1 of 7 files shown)


src/backend/app/Services/AbstractExportService.php
```diff
@@ -101,7 +102,8 @@ public function setActivatedSheet(string $sheetName): void {
         }
     }
 
-    public function saveSpreadSheet(): string {
+    public function saveSpreadSheet(?Filesystem $storage = null): string {
```
dmohns: 🫣 I'm not a big fan... It feels like we are giving tooooo many configuration options which has little gain, but adds large complexity. 

What do you think about the following idea: 

- `AbstractExportService.php` is the only thing that's really using `disk('local')`
- Let's change it, to use default, cloud storage here as well

Basically, this gets rid of the distinction between `disk('default')` vs `disk('local')`. This has the additional benefit of making Local Dev and Prod more similar.

beesaferoot:
Addressed your comments @dmohns  Kindly review.

PR #1206, opened 2025-12-16
Feat: Sort Customer List (ASC/DESC) 

munyanezaarmel:
Add customer list sorting by relevant fields
closes: #1117 
image

 I left phone, isActive, and device as NON-clickable because backend doesn't support sorting them yet

### Pull Request checklist

### DIFF
--- a/src/frontend/src/modules/Client/Clients.vue
+++ b/src/frontend/src/modules/Client/Clients.vue
@@ -372,6 +372,20 @@ export default {
         this.currentSortOrder = "desc"
       }
 
+      // Build term object with sort parameters for pagination
+      const term = {}
+      if (this.currentSortBy) {
+        const prefix = this.currentSortOrder === "desc" ? "-" : ""
+        term.sort_by = `${prefix}${this.currentSortBy}`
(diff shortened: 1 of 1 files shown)


beesaferoot (approved):

src/frontend/src/modules/Client/Clients.vue
```diff
@@ -698,4 +718,36 @@ export default {
...
+  }
+}
+
+::v-deep .md-table-head.md-sorted:hover .md-table-sortable-icon {
+  opacity: 1 !important;
+
+  svg {
+    fill: #448aff !important;
+  }
+}
```
dmohns: Uff... I find this very ugly 😅 Looks like it is introducing **a lot** of complexity. You are introducing this because you are fighting over some global styling in `SideBar.vue`, right? Could we instead fix the global styling in `Sidear.vue` first?

dmohns (approved):
Looks good now! Thanks a lot!

The issue with `SideBar.vue` messing with the styling persists. I will create a follow up ticket for that.

Issue #1118, opened 2025-11-18
[Feature Request]: Sort Transaction List (ASC/DESC)


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Similar to: https://github.com/EnAccess/micropowermanager/issues/1117

Currently, the transaction list is just displayed as-is with very limited possibility to interact.

<img width="1512" height="732" alt="Image" src="https://github.com/user-attachments/assets/52e87a72-0eca-47fc-bbec-177371597600" />

### Proposed Solution

Please implement sorting option ASC/DESC which is common in many other tools.

- Clicking the column header once, should sort asc by that column
- Clicking the column header again, should sort desc by that column
- Clicking the column header a third time should reset to default sorting

Please also adapt the UI such that the current sorting becomes clear to the user.

### Alternatives Considered

N/A

PR #1219, opened 2025-12-18
Sort Transaction List (ASC/DESC)

munyanezaarmel:
Sort Transaction List
closes: #1118

The sorting is already implemented in the codebase, but it’s not working in components like AgentTransactionList and others because they use the old/manual table structure with <md-table-row> and explicit loops. Vue Material’s sorting only works with v-model binding and scoped slots, so we need to migrate to the v-model approach.
### Pull Request checklist


dmohns:
Hey @munyanezaarmel any chance to add the changes from user list sorting to this PR as well?

beesaferoot:
> > Hey @beesaferoot! There's an issue with transaction sorting. When users click to sort the table:
> > 
> > * The frontend sends a sort_by parameter (like sort_by=-created_at)
> > * If there are any active filters, it calls /api/transactions/advanced
> > * That endpoint returns empty results or ignores the sorting
> >   The simple /api/transactions/list/all endpoint works fine with sorting, but /api/transactions/advanced doesn't seem to respect the sort_by parameter.
> >   Can you check the advanced endpoint to make sure it's handling sort_by correctly?
> 
> @munyanezaarmel Good catch. I don't think sorting was added to that endpoint. I will be looking into that.

Please check again, the endpoint now handles `sort_by` correctly. @munyanezaarmel

beesaferoot:
> Check let me know if it is working

Trying to test by selecting different sort fields (i.e clicking up and down arrows), it shows from the endpoint calls the frontend is doing something wrong here.

Screenshot 2026-01-21 at 8 24 07 PM

PR #1228, opened 2026-01-07
Remove hardcoded `inensus` namespace from plugins

dmohns:

Makes progress on: #1227 
Makes progress on: #1052

In the fist step we are changing the file structure of where Plugin (backend) code is stored. Before each plugin was considered a "package" (in composer terminology) and stored outside of `app`

```
├── app
├── database
├── ...
├── packages
│   └── inensus
│       ├── src
│       │   ├── Console
│       │   ├── Http
│       │   ├── Models
│       │   ├── Providers
│       │   └── Services
│       ├── .gitignore
│       ├── README.md
│       └── composer.json
└── composer.json
```

This is misleading, as these are no longer full-fledged packages. Their `composer.json` files were not used at all.

Instead we are changing this to

```
├── app
│   └── Plugins
│       ├── Console
│       ├── Http
│       ├── Models
│       ├── Providers
│       └── Services
├── database
├── ...
└── composer.json
```

Additionally, the Plugin creation `artisan` command was improved to automate some steps w.r.t. to the frontend code.
For example

- Automatically create a simple `Overview` page for new plugins and register it in `ExportedRoutes.js`

image

> [!WARNING]
> For now this change is only on file level. 

In particular the PHP path still include `namespace Inensus\OdysseyDataExport\Providers;`. This will be changed in a second step.

However, it means that we are currently violating PSR-4 auto loading conventions. We will accept this temporarily.

### Pull Request checklist

### DIFF
--- a/docs/development/plugins.md
+++ b/docs/development/plugins.md
@@ -18,7 +18,6 @@ Each plugin consists of:
 ## Quick Start
 
 1. **Prerequisites**
-
    - Running [advanced development environment](development-environment.md#advanced-development-environment) with local PHP installation
    - Basic knowledge of Laravel and Vue.js
 
@@ -34,14 +33,12 @@ Each plugin consists of:
    Replace `{plugin-name}` with the desired name of your plugin.
 
    This command:
-
    - Creates plugin backend code `src/backend/app/Plugins/{plugin-name}`
    - Creates plugin frontend code `src/frontend/src/plugins/{plugin-name}`
    - Automatically generates and registers a new plugin id
    - Automatically performs basic plugin registration (for example in frontend routes and backend plugins table)
 
 3. **Post-Creation Setup**
-
    - Review generated code structure
    - Follow integration steps below
 


beesaferoot (commented):
Left a few concerns above.

src/backend/app/Console/Commands/PluginGenerator.php
```diff
@@ -0,0 +1,312 @@
...
+        $this->outputComponents()->info('Creating plugin with following information:');
+
+        $this->outputComponents()->twoColumnDetail('Plugin name:', $pluginName);
+        $this->outputComponents()->twoColumnDetail('Plugin namespace:', $nameSpace);
+        $this->outputComponents()->twoColumnDetail('Description:', $description);
+
+        $this->outputComponents()->info('Running plugin generation script...');
+
+        // Determine project root
+        $projectRoot = File::isDirectory('/var/www/html') ? '/var/www/html' : base_path();
```
beesaferoot: Any reason we should bother about the path `/var/www/html`? the new-plugin command would ideally only be used during development.

PR #1241, opened 2026-01-14
Remove Inensus `namespace` and follow PSR-4 autoload convention

dmohns:

Closes: #1227 

### Pull Request checklist

### DIFF
--- a/src/backend/app/Console/Commands/PluginGenerator.php
+++ b/src/backend/app/Console/Commands/PluginGenerator.php
@@ -6,7 +6,9 @@ use Illuminate\Console\Command;
 use Illuminate\Support\Facades\File;
 
 class PluginGenerator extends Command {
-    protected $signature = 'micropowermanager:new-plugin {plugin-name} {--description= : Optional description for the plugin}';
+    protected $signature = 'micropowermanager:new-plugin
+        {plugin-name : Name of the new plugin. Use kebab-case (words separated by hyphens), e.g. `my-new-plugin`}
+        {--description= : Optional description for the plugin}';
     protected $description = 'Creates a new MicroPowerManager plugin';
 
     /**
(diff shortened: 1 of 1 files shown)


dmohns:
> We now seem to have an imposed plugin name constraint. Won't this cause issues? We don't enforce this during plugin creation, I think. Or alternatively highlight this on the docs.

Not sure I understand what you mean by constraint 🤔 The name **is** constrained in the sense that the plugin name (and namespace) must correspond with the file structure. The Plugin generator does some guessing and uses camel cases. The folder structure will automatically be correct: https://github.com/EnAccess/micropowermanager/pull/1241/files#diff-3579fde9d6895a20094bdb070d9d8136dbf739f5d6c77b18d95d67e3fd9c2ca2L208 

However, nothing is preventing you to call your Plugin `MyNeWpLuGiN`. But, you have to put it into `app/Plugins/MyNeWpLuGiN`.

In my opinion this is good practise, i.e. what PSR-4 enforced. Before we were violating that. Hence auto discovery didn't work and we had to manually add discovery rules. Which is tedious and confusing.

beesaferoot:
> > We now seem to have an imposed plugin name constraint. Won't this cause issues? We don't enforce this during plugin creation, I think. Or alternatively highlight this on the docs.
> 
> Not sure I understand what you mean by constraint 🤔 The name **is** constrained in the sense that the plugin name (and namespace) must correspond with the file structure. The Plugin generator does some guessing and uses camel cases. The folder structure will automatically be correct: https://github.com/EnAccess/micropowermanager/pull/1241/files#diff-3579fde9d6895a20094bdb070d9d8136dbf739f5d6c77b18d95d67e3fd9c2ca2L208
> 
> However, nothing is preventing you to call your Plugin `MyNeWpLuGiN`. But, you have to put it into `app/Plugins/MyNeWpLuGiN`.
> 
> In my opinion this is good practise, i.e. what PSR-4 enforced. Before we were violating that. Hence auto discovery didn't work and we had to manually add discovery rules. Which is tedious and confusing.

> Not sure I understand what you mean by constrain

I noticed files are being renamed to match the namespace i.e `App/Plugins/my-test-plugin` -> ``App/Plugins/MyTestPlugin` 

I might be missing an important aspect of PSR-4, but does it enforce the file names as well? My understanding is that it enforces the use of the correct qualified namespace. 

If indeed it does also enforce filenames (not very clear from the docs https://www.php-fig.org/psr/psr-4/) the output needs to be corrected.  

The file path to the plugin is not correct as `src/backend/app/Plugins/plugin-gateway` does not exists.   

```
  ⇂ Review the generated backend files in src/backend/app/Plugins/plugin-gateway  
  ⇂ Review the generated frontend files in src/frontend/src/plugins/plugin-gateway  
  ⇂ Run migration: `php artisan migrate`  
  ⇂ Start developing your new plugin. See the Plugin Development guide for more! https://micropowermanager.io/development/plugins.html ```

PR #1375, opened 2026-03-13
fix: add explicit compose network aliases for mysql and redis

### LINKED ISSUE #1374: Local development login fails because backend cannot resolve db in Docker Compose
## Summary

In local development, login with the seeded demo user can fail even with correct credentials because backend DB hostname resolution breaks inside Docker.

## Steps to reproduce

1. Start local dev stack with `docker compose up -d`.
2. Try to login from frontend using:
   - email: `demo_company_admin@example.com`
   - password: `123123`
3. Or call backend directly:
   - `POST /api/auth/login`

## Expected behavior

Login returns a JWT token.

## Actual behavior

Login fails with server-side errors similar to:

`SQLSTATE[HY000] [2002] php_network_getaddresses: getaddrinfo for db failed: Temporary failure in name resolution`

## Root cause

In this environment, service/container DNS aliases (`db`, and in some cases `redis`) were not consistently resolvable from backend containers.

## Proposed fix

Add explicit network aliases on the default compose network:

- `mysql`: aliases `db`, `mysql`
- `redis`: alias `redis`

Apply the same fix in:

- `docker-compose.yml`
- `docker-compose-prod.yml`
- `docker-compose-dockerhub.yml`

Then recreate containers with `docker compose up -d --force-recreate`.

Zhey-on:

This PR adds explicit Docker Compose default-network aliases for database/cache services to avoid DNS resolution failures in local and compose-based setups.

- `mysql` now has aliases: `db`, `mysql`
- `redis` now has alias: `redis`

Updated files:
- `docker-compose.yml`
- `docker-compose-prod.yml`
- `docker-compose-dockerhub.yml`

closes: #1374

No functional side effects expected. This only makes service discovery explicit and more reliable.

Manual validation:

1. Recreated the stack:
   - `docker compose up -d --force-recreate`
2. Verified service hostname resolution from backend container:
   - `db`, `mysql`, and `redis` now resolve to container IPs.
3. Verified backend auth endpoint with seeded demo credentials:
   - `POST /api/auth/login` with `demo_company_admin@example.com` / `123123`
   - received valid JWT token.
4. Verified frontend responds on `http://localhost:8001`.

### Pull Request checklist

### DIFF
--- a/docker-compose-dockerhub.yml
+++ b/docker-compose-dockerhub.yml
@@ -86,6 +86,10 @@ services:
       - redis_data:/data
     ports:
       - 6379:6379
+    networks:
+      default:
+        aliases:
+          - redis
 
   mysql:
     container_name: mysql
@@ -99,6 +103,11 @@ services:
       - mysql_data:/var/lib/mysql
     ports:
       - 3306:3306
+    networks:
+      default:
+        aliases:
+          - db
+          - mysql
     healthcheck:
       test: [CMD, mysqladmin, ping, -h, localhost]
       start_period: 10s
--- a/docker-compose-prod.yml
+++ b/docker-compose-prod.yml
@@ -90,6 +90,10 @@ services:
       - redis_data:/data
     ports:
       - 6379:6379
+    networks:
+      default:
+        aliases:
+          - redis
 
   mysql:
     container_name: mysql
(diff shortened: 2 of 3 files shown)


dmohns (changes requested):
Hello @Zhey-on 

This PR doesn't do anything. From https://docs.docker.com/reference/compose-file/networks/ each service should be discoverable by it's service name. You are adding aliases which **is** the service name, which is redundant.

Could you please provide the motivation for this change?

Issue #1232, opened 2026-01-12
Data Export/Import on MPM


beesaferoot:
### Overview

MPM needs a unified way to move data in and out of MPM into other platforms and also different versions of MPM.  

The challenge posed here is how to represent a data import/export layer as a flexible API that can satisfy the following requirements;

-  Ease of moving data from and into other CRM platforms ie. Upya, Angasa etc. Flexibility is a key component to make this smoother than it would normally be otherwise
- Ease of moving data from into two separate instances of MPM 
- Unified data view from an API perspective.

## Public Data Model   (external data transfer i.e Upya, Angaza)
- Clusters i.e Site, Mini-grid, Village.
- Devices ie. Meters, Solar Home Systems, Ebikes.
- Customers i.e Address, customer type, customer group. 
- Transactions i.e Provider
- Appliance i.e Appliance type, Appliance SKUs, Payment Plan (Rates) (units)

## MPM instance Data Model 

- Users i.e Permisions, Admin. 
- MPM settings 
- All Public Data Models 

## Data format 

JSON (used strictly for MPM to MPM migration)
CSV
excel

PR #1377, opened 2026-03-13
Data Import/Export API: Extend import API

beesaferoot:
Makes progress on #1232

### Pull Request checklist


dmohns:
I'm trying to modify an existing user. It works, but the confirmation dialog looks like this:

image

Same when adding a new Device. The device gets added. But the dialog shows 0/0/0.

dmohns:
Is there a way to surface the import error to the users?

For example

image

is not very helpful.

But what I can see in dev tools is

image

beesaferoot:
> Is there a way to surface the import error to the users?
> 
> For example
> 
> image
> is not very helpful.
> 
> But what I can see in dev tools is
> 
> image

There should be a way, but it will be quite tricky. Handling bulk uploads, the errors can become pretty large. 

How about adding this as an improvement on subsequent PRs?

beesaferoot:
> I'm trying to modify an existing user. It works, but the confirmation dialog looks like this:
> 
> image
> Same when adding a new Device. The device gets added. But the dialog shows 0/0/0.

This has now been fixed.

dmohns (approved):
Nice! Let's keep the error handling as a future improvement.

Issue #1243, opened 2026-01-15
[Feature Request]: Transaction Abstraction Refactoring


beesaferoot:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

The current MPM transaction system functions reliably in most cases, but over time it has accumulated architectural constraints that make it difficult to maintain, reason about, and extend. These issues increasingly surface as the system grows in complexity and transaction volume.

### 1. Complex Model Relationships

The transaction domain relies on deeply coupled and highly interconnected models, making the system hard to understand and modify safely.

<img width="2039" height="1590" alt="Image" src="https://github.com/user-attachments/assets/47b23831-d8e1-48c2-9a51-605fd0a55596" />

Impact:
High cognitive overhead for contributors, increased risk of regressions, and limited flexibility for future changes.

### 2. Inconsistent Transaction Flows

Different transaction types follow significantly different execution paths.
CashTransaction, for example, operates under a markedly different flow compared to other transaction types. The rationale for this divergence is unclear, and it is not obvious whether this approach provides any architectural or operational advantage.

Impact:
Increased complexity, duplicated logic, and difficulty enforcing consistent guarantees across transaction types.

### 3. Limited Retry Support

Retry logic is implemented only in TokenProcessor (3 attempts with a fixed 5-second delay). Other critical stages do not support retries:

- Validation failures → immediate failure

- Database errors during payment distribution → permanent failure

- SMS notification failures → silent failure

Impact:
Transient issues (network timeouts, rate limits, temporary database errors) result in irreversible transaction failures, often requiring manual intervention.

### 4. Implicit and Fragmented State Management

Transaction state is spread across multiple fields and models, with no single authoritative source of truth:

- Transaction.type is mutated during processing

- Provider-specific status codes vary (1 = success, 0 = pending, -1 = failed)

- No explicit tracking of processing stages or failure points

Impact:
It is difficult to:

- Identify “stuck” or partially processed transactions

- Resume processing from the point of failure

- Build reliable monitoring or observability dashboards

[...]

PR #1385, opened 2026-03-18
Refactor: Generialize payment initiation

beesaferoot:
This changes attempts to consolidate Payment initiation in on central location PaymentInitialization. It also introduces another payment method with sold appliance.

Makes progress on #1243

Screenshot 2026-03-18 at 3 05 40 PM

### Pull Request checklist


docs/usage-guide/images/payment-flow.svg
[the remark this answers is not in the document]
beesaferoot: I was referencing to validations done on `initializePayment`  for example with Paystack provider which goes through a call chain that ends up in `InitializeTransactionResource` which does validation on the structure of the data. 

https://github.com/EnAccess/micropowermanager/pull/1385/changes#diff-d52a331b98749532774430e83f1cbed2cce947a64d05307ad552ece71ff25c4bR156

docs/usage-guide/images/payment-flow.excalidraw.svg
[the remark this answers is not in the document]
beesaferoot: Thanks for sharing.

dmohns (commented):
I'm still struggling a little bit to follow the payment processing flow. Both, with and without your changes 😅 

After coming back to this multiple times, I think one of the confusions for me lies in the redundancy between Transaction **Providers** (Providers here mean Laravel Providers, not Payment Providers) and Transaction **Services** 

For Providers we already have a Contract `ITransactionProvider`. For Services we have `AbstractPaymentAggregatorTransactionService` 

Now, looking at `ITransactionProvider` it seems like it was intended to already cover a lot of aspects that you are trying to add here. 

For example it has 

```php
public function init(BasePaymentProviderTransaction $transaction): void;
```

Do you think it within the scope of this PR to clean up this redundancy?

beesaferoot:
> I'm still struggling a little bit to follow the payment processing flow. Both, with and without your changes 😅
> 
> After coming back to this multiple times, I think one of the confusions for me lies in the redundancy between Transaction **Providers** (Providers here mean Laravel Providers, not Payment Providers) and Transaction **Services**
> 
> For Providers we already have a Contract `ITransactionProvider`. For Services we have `AbstractPaymentAggregatorTransactionService`
> 
> Now, looking at `ITransactionProvider` it seems like it was intended to already cover a lot of aspects that you are trying to add here.
> 
> For example it has
> 
> ```
> public function init(BasePaymentProviderTransaction $transaction): void;
> ```
> 
> Do you think it within the scope of this PR to clean up this redundancy?


If I understand your comment correctly `ITransactionProvider` should essentially also cover what `PaymentInitializer` is trying do to here. 

> Do you think it within the scope of this PR to clean up this redundancy?

Yes, this is very much part of the chain of refactor we are ultimately aiming to achieve. 

> I'm still struggling a little bit to follow the payment processing flow. Both, with and without your changes 😅

To be fair, this only PR tries to make it possible to bring all payment initiation steps in one place. My hope is that this will open up better abstractions that could simplify the entire flow, which is still rather complex.

dmohns:
Hey @beesaferoot coming back to this one. Can you resolve merge conflicts?

src/backend/app/Http/Controllers/AppliancePaymentController.php
```diff
@@ -40,32 +51,52 @@ public function checkStatus(int $transactionId): ApiResource {
...
+            sender: $sender,
+            message: $message,
+            type: Transaction::TYPE_DEFERRED_PAYMENT,
+            customerId: $applianceOwner->id,
+            serialId: $deviceSerial,
+        );
+
+        if ($providerId === $this::CASH_TRANSACTION_PROVIVER) {
+            dispatch(new ProcessPayment($companyId, $result['transaction']->id));
+        }
```
dmohns, line 94: Having this here feel odd. Can we move this into `CashTransactionService::initializePayment(...)`. I.e. for a CashTransaction if it's get "initialised" it get's immediate processed. By design of cash transaction

src/backend/app/Http/Controllers/AppliancePaymentController.php
```diff
@@ -40,32 +51,52 @@ public function checkStatus(int $transactionId): ApiResource {
...
+
+        return ApiResource::make($providers);
+    }
+
     /**
      * @return array<string, mixed>
      */
-    public function getPaymentForAppliance(Request $request, AppliancePerson $appliancePerson): array {
-        $creatorId = auth('api')->user()->id;
+    private function getPaymentForAppliance(Request $request, AppliancePerson $appliancePerson): array {
```
dmohns: Can we find a better name for this function that represents better what it actually does?

src/backend/app/Http/Controllers/AppliancePaymentController.php
```diff
@@ -40,32 +51,52 @@ public function checkStatus(int $transactionId): ApiResource {
...
         $amount = (float) $request->input('amount');
+        $providerId = (int) $request->input('payment_provider', 0);
```
dmohns:
```suggestion
        $providerId = $request->integer('payment_provider', 0);
```

😂

src/backend/app/Http/Controllers/AppliancePersonController.php
```diff
@@ -133,31 +138,47 @@ private function assignDevice(AppliancePerson $appliancePerson, Request $request
...
-            payer: $appliancePerson->person,
-            transaction: $transaction,
-        ));
-        event(new TransactionSuccessfulEvent($transaction));
+        if ($paymentProviderId === $this::CASH_TRANSACTION_PROVIVER) {
+            $applianceRate = $this->applianceRateService->createPaidRate($appliancePerson, $downPayment);
+            event(new PaymentSuccessEvent(
+                amount: (int) $result['transaction']->amount,
+                paymentService: 'web',
+                paymentType: 'down payment',
```
dmohns: Could we use `Transaction::TYPE_DOWN_PAYMENT` here?

dmohns: Note this will change from `down payment` to `down_payment`. I think we will need a migration in this case?

dmohns (commented):
Couple of comments, hope they make sense

src/backend/app/Http/Controllers/AppliancePersonController.php
beesaferoot: Marking this a future improvement.

beesaferoot: Actually this works because payment type is just a varchar type

dmohns: Does this mean, it won't get persisted into database?

beesaferoot: It will get persisted, but it doesn't need a migration.

beesaferoot: We handle both cases on the UI side: 

```js
  formatPaymentType(type) {
      const labels = {
        energy_service: this.$tc("phrases.eaasRate"),
        eaas_rate: this.$tc("phrases.eaasRate"),
        "down payment": this.$tc("phrases.downPayment"),
        down_payment: this.$tc("phrases.downPayment"),
        installment: this.$tc("phrases.deferredPayment"),
        deferred_payment: this.$tc("phrases.deferredPayment"),
        energy: this.$tc("words.energy"),
        access_rate: this.$tc("phrases.accessRate"),
      }
      return labels[type] || type
    },
```

dmohns:
LGTM, can you fix the `larastan` then I can approve

dmohns (approved):
Nice!!! Let's go with this

PR #1438, opened 2026-04-14
Add `pt` language

dmohns:

Fully Claude generated `pt` translations. We consider this a first step to kick start adoption of MPM in Mozambique. In a later step we should review these translations w.r.t. Off Grid terminology in MZ Portuguese. 

To be confirmed if this is the correct Mozambican/European Portuguese flavor (e.g., "contador" for meter)

### Pull Request checklist


tools/sort_lang_json.sh
```diff
@@ -0,0 +1,32 @@
...
+for file in "$DIR"/*.json; do
+  # Skip if no json files are found
+  [ -e "$file" ] || continue
+
+  echo "Sorting $file ..."
+  tmpfile=$(mktemp)
+  jq -S . "$file" > "$tmpfile" && mv "$tmpfile" "$file"
+done
+
+echo "✅ All JSON files in '$DIR' have been sorted."
```
beesaferoot, line 32: Can we do this in PHP?

beesaferoot (commented):
Left a comment.

tools/sort_lang_json.sh
dmohns: Hmm... This only touches frontend files and has nothing to do with the BE at all. I would find it odd to write an `artisan` command to do this. 

I do get it that `sh` is not the most inclusive. Maybe I could see an argument to add this as a `node` script? But then we should add `check` and `fix` scripts as well. Probably this is a future improvement tho. WDYT?

beesaferoot: Your point about being unrelated to BE - I agree with. Adding an npm script (via `package.json`)  on the `frontend` folder would be better suited. 

My concern is that we are creating a `tools` folder outside the frontend source code to run a bash script.

beesaferoot: I'm fine moving this to a differen  improvement issue.

beesaferoot (approved):

PR #1512, opened 2026-06-11
Add has crud operations trait

dmohns:

This PR add `HasCrudOperations` trait and remove more code repetition.

Technically, we are introducing a breaking change here. the 4 services

- `CompanyService`
- `GeographicalInformationService`
- `MainSettingsService`
- `RegistrationTailService`

Each declared `getAll(?int $limit = null)` but always returned `->get()`, ignoring the limit.

But it looks like it wasn't used in the code.

### Pull Request checklist

### DIFF
--- a/src/backend/app/Plugins/PaystackPaymentProvider/Services/PaystackTransactionService.php
+++ b/src/backend/app/Plugins/PaystackPaymentProvider/Services/PaystackTransactionService.php
@@ -17,7 +17,6 @@ use App\Services\Interfaces\IBaseService;
 use App\Services\Interfaces\PaymentInitiator;
 use App\Services\PersonService;
 use Illuminate\Database\Eloquent\Collection;
-use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Support\Facades\DB;
 use Ramsey\Uuid\Uuid;
 
--- a/src/backend/app/Plugins/PaystackPaymentProvider/Services/PaystackTransactionService.php
+++ b/src/backend/app/Plugins/PaystackPaymentProvider/Services/PaystackTransactionService.php
@@ -22,6 +22,8 @@ use Illuminate\Support\Facades\DB;
 use Ramsey\Uuid\Uuid;
 
 /**
+ * @extends AbstractPaymentAggregatorTransactionService<PaystackTransaction>
+ *
  * @implements IBaseService<PaystackTransaction>
  */
 class PaystackTransactionService extends AbstractPaymentAggregatorTransactionService implements IBaseService, PaymentInitiator {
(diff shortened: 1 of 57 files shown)


src/backend/app/Plugins/PaystackPaymentProvider/Services/PaystackTransactionService.php
```diff
@@ -85,13 +86,6 @@ public function getById(int $id): ?PaystackTransaction {
         return $this->paystackTransaction->newQuery()->find($id);
     }
 
-    public function update($paystackTransaction, array $paystackTransactionData): PaystackTransaction {
-        $paystackTransaction->update($paystackTransactionData);
-        $paystackTransaction->fresh();
-
-        return $paystackTransaction;
-    }
-
```
beesaferoot, line 94: `update`, `getAll` and `delete` methods where removed but I can't find any were the implementation can be called from. Did you mean to use the `HasCrudOperations` trait here as well?

beesaferoot (commented):
Keep the improvement PRs coming 💪🏼

I left some comments

beesaferoot:
Neat pick: Most of the applications of `HasCrudOperations` essentially satifish `IBaseService` interface. 

Could we consider use an abstract class i.e 

```php

/**
 * @template T of Model
 * @implements IBaseService<T>
 */
abstract class BaseService implements IBaseService
{
    /**
     * @return T
     */
    abstract protected function model(): Model;

    /**
     * @return T|null
     */
    public function getById(int $id): ?Model
    {
        return $this->model()->newQuery()->find($id);
    }

    // create(), update(), delete(), getAll()
}

// can use reference like this 
class CityService extends BaseService
{
}
```

src/backend/app/Plugins/PaystackPaymentProvider/Services/PaystackTransactionService.php
dmohns: ~~Ohhh, yes I did mean to add it. Good catch~~

dmohns:
> Could be consider use an abstract class i.e

Yeah, I did consider that. This is a bit of a subjective topic. I, personally, prefer composition over inheritance as it better resembles how I think about the code 😅 Our code base in itself is not 100% consistent. However, in Laravel there seems to be a tendency to use traits (composition) more and more. Essentially, here I want to try this pattern out 😂 

Hope this makes sense?

src/backend/app/Plugins/PaystackPaymentProvider/Services/PaystackTransactionService.php
dmohns: Wait... no. The base class `AbstractPaymentAggregatorTransactionService` has the `HasCrudOperations` trait. Hence `PaystackTransactionService` inherits it.

beesaferoot: Okay, I missed that. Thanks for pointing out

beesaferoot:
> > Could be consider use an abstract class i.e
> 
> Yeah, I did consider that. This is a bit of a subjective topic. I, personally, prefer composition over inheritance as it better resembles how I think about the code 😅 Our code base in itself is not 100% consistent. However, in Laravel there seems to be a tendency to use traits (composition) more and more. Essentially, here I want to try this pattern out 😂
> 
> Hope this makes sense?

Yeah, I suppose no harm is going one direction or the other.  One other question i had was to confirm if it was enforceable on the codebase?  if a class is composed of the methods from the trait and does not include define `crudModel` do we flag that on the pipeline?

dmohns:
> if a class is composed of the methods from the trait and does not include define `crudModel` do we flag that on the pipeline?

Yes! Because `crudModel` has the `abstract` keyword. So, using the trait forces you to implement it.

beesaferoot (approved):

PR #1516, opened 2026-06-11
Use PHP8.4's asymmetric visibility

dmohns:

Using PHP8.4's asymmetric visibility to improve repetition and boilerplate in our code base. https://www.php.net/manual/en/language.oop5.visibility.php

As an example, applying this to `src/backend/app/Services/AbstractPaymentAggregatorTransactionService.php`. If we like this, we could expand this to the broader code base.

### Pull Request checklist


beesaferoot (commented):
My major concern is the runtime guarantees. If we can't guarantee wrong usage that looks seemingly okay doesn't blow up at runtime, then I will be bought. Otherwise, I would opt for the original verbose approach.

dmohns:
> My major concern is the runtime guarantees. If we can't guarantee wrong usage that looks seemingly okay doesn't blow up at runtime, then I will be bought. Otherwise, I would opt for the original verbose approach.

I'm playing around with it to confirm the behaviour.

image

Larastan complains about it: 

```
 ------ ----------------------------------------------------------------------------------------------------------------------------------- 
  Line   app/Plugins/SwiftaPaymentProvider/Providers/SwiftaTransactionProvider.php                                                          
 ------ -----------------------------------------------------------------------------------------------------------------------------------                                              
  :89    Assign to private(set) property App\Plugins\SwiftaPaymentProvider\Services\SwiftaTransactionService::$paymentProviderTransaction.  
         🪪  assign.propertyPrivateSet                                                                                                      
 ------ ----------------------------------------------------------------------------------------------------------------------------------- 
```

From what I understand the behaviour is similar, as if you'd try to access a `private`/`protected` property from out of scope. I think this is intuitive behaviour and shouldn't slip into prod. Assuming we run Larastan every time, which we do.

beesaferoot:
> > My major concern is the runtime guarantees. If we can't guarantee wrong usage that looks seemingly okay doesn't blow up at runtime, then I will be bought. Otherwise, I would opt for the original verbose approach.
> 
> I'm playing around with it to confirm the behaviour.
> 
> image
> Larastan complains about it:
> 
> ```
>  ------ ----------------------------------------------------------------------------------------------------------------------------------- 
>   Line   app/Plugins/SwiftaPaymentProvider/Providers/SwiftaTransactionProvider.php                                                          
>  ------ -----------------------------------------------------------------------------------------------------------------------------------                                              
>   :89    Assign to private(set) property App\Plugins\SwiftaPaymentProvider\Services\SwiftaTransactionService::$paymentProviderTransaction.  
>          🪪  assign.propertyPrivateSet                                                                                                      
>  ------ ----------------------------------------------------------------------------------------------------------------------------------- 
> ```
> 
> From what I understand the behaviour is similar, as if you'd try to access a `private`/`protected` property from out of scope. I think this is intuitive behaviour and shouldn't slip into prod. Assuming we run Larastan every time, which we do.

Nice, then it looks that we have a way of confirming it is used appropriately before getting merged.

beesaferoot (approved):

PR #1536, opened 2026-06-26
refactor: migrate GeoInfo points filed into GeoJSON data scheme via `geo_json` field

### LINKED ISSUE #1035: [Feature Request]: Convert `GeographicalInformation` `point` to `GeoJSON`
### Preflight Checklist

### Problem Description

Currently, `GeographicalInformation` is not very flexible as we store only points.

### Proposed Solution

Convert `GeographicalInformation` to store information as `GeoJSON`

- Add a new field to `GeographicalInformation` called `geo_json`
- Create a migration that populates `geo_json` with 
    ```json
    {
      "type": "Point",
      "coordinates": [ ..., ... ]
    }
    ```
- Drop `points`
- Change all controllers to expose a) lat,lon for backwards compatability b) the newly created geo_json

### Alternatives Considered

N/A

### Additional Information

_No response_

beesaferoot:
Ths PR carries out a large but crucial refactor before landing MPM v1.0. This continues a string of refactors that replaces changes existing location data into on consistent standard GeoJSON format. 

Closes #1035 

### Pull Request checklist

### DIFF
--- a/src/backend/app/Models/GeographicalInformation.php
+++ b/src/backend/app/Models/GeographicalInformation.php
@@ -15,7 +15,7 @@ use Illuminate\Support\Carbon;
  * @property      int         $id
  * @property      int         $owner_id
  * @property      string      $owner_type
- * @property      string      $points
+ * @property      object|null $geo_json
  * @property      Carbon|null $created_at
  * @property      Carbon|null $updated_at
  * @property-read Model       $owner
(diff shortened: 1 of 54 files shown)


src/backend/app/Models/GeographicalInformation.php
```diff
@@ -32,4 +32,63 @@ class GeographicalInformation extends BaseModel {
...
+    public static function makePoint(float $latitude, float $longitude): object {
+        return (object) [
+            'type' => 'Feature',
+            'geometry' => (object) [
+                'type' => 'Point',
+                'coordinates' => [$longitude, $latitude],
+            ],
+            'properties' => new \stdClass(),
+        ];
+    }
```
dmohns: I'm not a big fan of "manually" writing a `GeoJSON`. As part of this refactor, could we evaluate using a library to do that? From a quick research https://github.com/brick/geo seems to be a promising candidate.

src/backend/app/Models/GeographicalInformation.php
```diff
@@ -32,4 +32,63 @@ class GeographicalInformation extends BaseModel {
...
+        return [
+            'geo_json' => 'object',
+        ];
+    }
+
+    /**
+     * Build a GeoJSON Point Feature from a latitude/longitude pair. The shape mirrors the
+     * Cluster model's `geo_json` (a Feature with `geometry` + `properties`) so the whole codebase
+     * has one GeoJSON convention. GeoJSON orders coordinates as [longitude, latitude] — the reverse
+     * of how humans say them — and this is the single place that ordering is encoded.
```
dmohns: Can you write this comment with a bit more "matter of fact" tone? I.e. GeoJSON uses [RFC 7946](https://datatracker.ietf.org/doc/html/rfc7946) (long / lat format) while most humans are more familiar with [ISO 6709](https://en.wikipedia.org/wiki/ISO_6709)

dmohns (commented):
Really nice!! Looking forward to it. Few comments

dmohns:
Nice!!! Last question... can you expand why you chose `jmikola/geojson`? It seems a bit inactive, last commit 3 years ago, only tested against PHP 8.2. Compared to the more active `brick/geo` ?

beesaferoot:
> Nice!!! Last question... can you expand why you chose `jmikola/geojson`? It seems a bit inactive, last commit 3 years ago, only tested against PHP 8.2. Compared to the more active `brick/geo` ?

 `brick/geo` comes with a ton of features we don't utilize more than constructing a GeoJSON object i.e performing operations on these objects. While `jmikola/geojson` implements the GeoJSON format specification and doesn't add anything else; which I think is just right for our use case.   

I didn't think we needed a dependency that we don't use 99% of its features.

dmohns (approved):

PR #1540, opened 2026-06-30
Add Scramble Extension for `ApiResource`

dmohns:

This PR adds a purely vibe-coded Scramble extension to properly parse `return ApiResource::make( ... )` calls in controllers. We aim to support both case of keyed-arrays like

```php
        return ApiResource::make([
            'transaction_id' => $transaction->id,
            'token' => $transaction->token,
        ]);
```

and collections

```php
return ApiResource::make($this->agentTransactionService->getByCustomerId($agent->id, $customerId));
```

**Caveat:**

We cannot properly resolve _nested_ returns. For example the People Endpoint returning nested `addresses` and `devices`

image

In this case it would be even more beneficial to move to proper Resources. But it is left as a future improvement.

**Before (keyed-array):**

image

**Before (collection):**

image

**After (keyed-array):**

image

**After (collection):**

image

### Pull Request checklist


beesaferoot:
This looks really good. Great work @dmohns

beesaferoot (approved):

Issue #743, opened 2025-06-12
[Feature Request]: Sending on/off commands to a given Calinmeter


tushabe:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

As a microgrid developer, I would like to use MPM to remotely turn on/off powerflow for CalinSmartMeter so that I can steer energy usage on the microgrid

### Proposed Solution

Add a toggle on/off switch in the MPM UI for each meter that toggles builtin meter relay on/off

### Alternatives Considered

My current alternative is to connect to the CalinSmartMeters via their proprietary AMI and toggle them on/off that way. I can show you a demo for this in a week or two when we have some CalinSmartMeters connected. 

### Additional Information

They provided me [documentation](https://www.postman.com/nearly-free-energy/nearly-free-energy/overview) for their API (postman collection).

dmohns:
Hey @tushabe 

thanks for this Feature Request. We have received similar requests in the past, but never documented them as well.

I have a clarification request here: As a user, could you describe a bit more in detail "where in the UI" you would expect to find this functionality? On the Meter detail page? On the User page?

tushabe:
Got this UML diagram out of the code

![Image](https://github.com/user-attachments/assets/0b867222-7b7f-4169-bf9e-d9a90b3cc458)

tushabe:
@dmohns I think I have an answer for you, this project is really well maintained so reviewing docs and demo were really quicker than I expected. Here's my thinking. 

This action belongs with the Meter not the User so it's definitely makes sense to me to have on the Meter details page and at least a "Relay status" indicator on the meters overview list page. Let me try to mock it up. 

┌─────────────────────────────────────────────────┐
│ Meter Details: [Meter Serial #]                 │
├─────────────────────────────────────────────────┤
│ General Info | Readings | Configuration | Actions│
├─────────────────────────────────────────────────┤
│                                                 │
│ Connection Status: ● Connected                  │
│                                                 │
│ Relay Status: ON                                │
│                                                 │
[...]

Issue #490, opened 2025-02-19
[Feature Request]: Manual maintenance for individual SHS via Tokens


dmohns:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/EnAccess/micropowermanager/blob/main/CONTRIBUTING.md) for this project, if it exists.
- [x] I agree to follow the [Code of Conduct](https://github.com/EnAccess/micropowermanager/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/EnAccess/micropowermanager/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

In BAU operation of a SHS distributor it is sometimes required to perform certain PayGO actions an specific devices. This include (but are not limited to)

- Enabling a device for a certain time period
- Disabling a device
- Reset a devices token counter

### Proposed Solution

In the MPM Device overview it should be possible to generate Tokens for the above mentioned use-cases for the operator.

NB: The delivery mechanism has to be decided, i.e. if those generated token should also be shipped to customer directly via SMS or "jus" displayed on the website.

### Alternatives Considered

N/A

PR #1543, opened 2026-07-01
feature: verify device mapping using manufacturer query api

beesaferoot:
IManufacturerDeviceInfo interface is added to compose a Manufacturer class with device info retrieval logic

### Pull Request checklist


dmohns:
I'm not sure I like the UX here (or, I don't understand fully understand the use case). Right now, we are simply adding a button. It's not really clear to me, under which circumstances a user would ever click that.

What I can imagine is we are looking for some way to confirm we can actually control a given device from MPM. But if this is the case I would consider a different approach:

- On the device model we could add two fields liked `mapped_with_manufacturer_api` (or similar) and `last_checked`. - We could then nicely add these information to a Device Detail overview. Also, Device Overview and search filter + export
- When new devices are added, we immediately run the check.
- On Device Details view, the button you added would run the check as above.

In terms of API design another questions... We have a couple of feature requests regarding ad-hoc device management, for example

- https://github.com/EnAccess/micropowermanager/issues/490
- https://github.com/EnAccess/micropowermanager/issues/743

I imagine there will be some overlap with the work you are doing here. Have you considered if there will be some synergy between the two works streams?

beesaferoot:
> I'm not sure I like the UX here (or, I don't understand fully understand the use case). Right now, we are simply adding a button. It's not really clear to me, under which circumstances a user would ever click that.
> 
> What I can imagine is we are looking for some way to confirm we can actually control a given device from MPM. But if this is the case I would consider a different approach:
> 
> * On the device model we could add two fields liked `mapped_with_manufacturer_api` (or similar) and `last_checked`. - We could then nicely add these information to a Device Detail overview. Also, Device Overview and search filter + export
> * When new devices are added, we immediately run the check.
> * On Device Details view, the button you added would run the check as above.
> 
> In terms of API design another questions... We have a couple of feature requests regarding ad-hoc device management, for example
> 
> * [[Feature Request]: Manual maintenance for individual SHS via Tokens #490](https://github.com/EnAccess/micropowermanager/issues/490)
> * [[Feature Request]: Sending on/off commands to a given Calinmeter #743](https://github.com/EnAccess/micropowermanager/issues/743)
> 
> I imagine there will be some overlap with the work you are doing here. Have you considered if there will be some synergy between the two works streams?

You make good points; I will be looking further into it with your feedbacks. Thanks

dmohns (commented):
This look very good 👍 Only slight concern I have is about naming

- There is some in consistency, sometime you are referring to it as ManufacturerMapping sometimes as DeviceMapping. Could we decide on on terminology and stick with it?
- I personally find the word "mapping" not specific enough, it's more of an implementation detail than descriptive (i.e. some manufacturer say a device is mapped to an API key but other manufacturer might totally avoid this naming). I'm not really sure what would be better 🫣 Sometime along the lines of device management? Or Remote device management? Any ideas?

beesaferoot:
> This look very good 👍 Only slight concern I have is about naming
> 
> * There is some in consistency, sometime you are referring to it as ManufacturerMapping sometimes as DeviceMapping. Could we decide on on terminology and stick with it?
> * I personally find the word "mapping" not specific enough, it's more of an implementation detail than descriptive (i.e. some manufacturer say a device is mapped to an API key but other manufacturer might totally avoid this naming). I'm not really sure what would be better 🫣 Sometime along the lines of device management? Or Remote device management? Any ideas?

Device mapping here only represents the verification on the manufacturer level. Device Management was the intended name for the abstraction that includes work from #490, #743 but left it at mapping on this PR. 

I could make it uniform by using "Device management" for the section header already.

dmohns:
> Device mapping here only represents the verification on the manufacturer level. Device Management was the intended name for the abstraction that includes work from #490, #743 but left it at mapping on this PR.
> 
> I could make it uniform by using "Device management" for the section header already.

Yeah, I understand what you mean with "Device Management" and it makes sense. This can be expanded later. 

My point is, what we are currently calling "mapping" is really checking if we have permissions to call the management API (or remote management permission). Mapping generally refers to combine or link to entities, which doesn't make sense from our point of view (It _does_ make sense from the manufacturer point of view, they are mapping an API key to a device, which gives it permissions).

But maybe I'm being way too nit-picky here 🫣

beesaferoot:
> > Device mapping here only represents the verification on the manufacturer level. Device Management was the intended name for the abstraction that includes work from #490, #743 but left it at mapping on this PR.
> > I could make it uniform by using "Device management" for the section header already.
> 
> Yeah, I understand what you mean with "Device Management" and it makes sense. This can be expanded later.
> 
> My point is, what we are currently calling "mapping" is really checking if we have permissions to call the management API (or remote management permission). Mapping generally refers to combine or link to entities, which doesn't make sense from our point of view (It _does_ make sense from the manufacturer point of view, they are mapping an API key to a device, which gives it permissions).
> 
> But maybe I'm being way too nit-picky here 🫣

For me Mapping here simply implies that the device unit exists on the Manufacturer platform which then allows for operations on that unit i.e unlocking, credit unit, device control etc.  

It doesn't care about the means that we verify this on the manufacturer side. If the unit gets removed on the Manufacturer side, then we can correctly flag this as such using the button. 


We can go with this for now and refactor to a better name if a better suggestion surfaces.

dmohns (approved):

PR #1546, opened 2026-07-02
refactor: Import endpoint utilize DTO pattern

dmohns:

This PR adds refactoring to the Import endpoints aiming to making the code base more idiomatic and (as a result) provide a better auto generated documentation.

- Utilize specific resources for returns.
- Use DTO instead of untyped arrays
- Exception handling moved to dedicated exception

**Before:**

image

**After:**

image

> [!INFO]
> This PR is an intermediate steps. In a second step I suggest to move to Laravel Data for even more idiomatic code organisation.
> However, implementing this in a second PR.

⚠️ Breaking change: import endpoints accept one canonical payload format

### What changed

All `POST /api/import/*` endpoints now accept exactly **one** request format — the `data` key
must be a **list of items to import**:

```json
{ "data": [ { "device_info": { "serial_number": "SM-001" } }, ... ] }
```

Previously, the endpoints silently tolerated a second, undocumented format: an entire exported
JSON file posted under data (a "double-wrapped" payload):

```json
{ "data": { "data": [ ... ], "meta": { "exported_at": "..." } } }
```

### Pull Request checklist

### DIFF
--- a/collections/Advanced Transaction Search.bru
+++ b/collections/Advanced Transaction Search.bru
@@ -1,7 +1,7 @@
 meta {
   name: Advanced Transaction Search
   type: http
-  seq: 34
+  seq: 39
 }
 
 get {
--- a/collections/Check Login (API).bru
+++ b/collections/Check Login (API).bru
@@ -1,7 +1,7 @@
 meta {
   name: Check Login (API)
   type: http
-  seq: 12
+  seq: 13
 }
 
 get {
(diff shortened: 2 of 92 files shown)


beesaferoot (approved):
Looks really good. Tested  locally and not major breaking change was observed. 

I noticed that we don't add the geojson data of mini-grids when we export so it shows that the mini-grid has no location when a cluster is imported.

PR #1572, opened 2026-07-20
Fix/agent usability

### LINKED ISSUE #1569: [Bug]: Agents Transactions getting duplicated
### Preflight Checklist

### Expected Behavior

When an agents collects a payment, it should be recorded once and not affect the balance.

### Actual Behavior

Whenever an agent collects a payment, the transaction gets duplicated and it makes the agents balance higher than it should be. 
For example, if agent collects 100, in the transaction history we find 2 transactions for the same amount.

Image

### Steps to reproduce

Create agent, choose agent commission type, assign units, go to field App, make payment for a customer.

Windows

### Additional Information

_No response_

beesaferoot:
Close: #1569 

This PR addresses a number Agent usability issues. 

### Pull Request checklist


docs/usage-guide/agent-users.md
```diff
@@ -72,7 +81,9 @@ These records track money given to and returned by agents, keeping their balance
 
 ### Agent Charges
 
-An **agent charge** represents money the company gives to the agent so that they can continue serving customers. Charges are created from the web panel (`Agents` → `Commission Types` → `+`). When a charge is saved, a matching balance entry is added so the credit appears in the agent’s ledger.
+An **agent charge** represents money the company gives to the agent so that they can continue serving customers.
```
dmohns: I still do not understand why a company would need to do that, but probably not a consideration for this change 😂

docs/usage-guide/agent-users.md
```diff
@@ -18,13 +18,19 @@ For more information on how to generate and manage tickets, see [Tickets](/usage
 ![Agent Commission Types](images/agent-commission-types.png)
 
 Agents receive cash from customers on site.
-An agent's **outstanding balance** is the money it owes the company (i.e. it is the money it has collected on behalf of the company and has to transfer accordingly).
+MicroPowerManager tracks three separate totals for every agent, all visible on the agent's profile card:
+
+- **Balance**: the agent's running position.
+  It goes negative as the agent sells on the company's behalf and positive when the company pre-loads credit or the agent settles up.
```
dmohns: Selling something makes the balance go negative? I cannot help but finding this extremely unintuitive 🙈 

Is there anyway we can refactor these values to behave in the naive way? i.e. balance representing the amount of money an agent holds?

beesaferoot: > Is there anyway we can refactor these values to behave in the naive way? i.e. balance representing the amount of money an agent holds?

Definitely, I could look into this on the PR.

beesaferoot: I went over some of the terminology, and I made some refactoring to improve the terminology for now.

dmohns: Thanks for looking into that 🙏 Unfortunately (for me) it doesn't change much. We are "just" adding new terminology. 

Look what's happening: We need a lot of usage guide to explain what all the different terms mean. Then, everywhere in the code we need to explain the reader: this record is negative, this record is positive. 

What I have in mind is the following:

- There is only one Agent Balance number. This number represent the amount company money the agent holds.

To me, this is the most intuitive. Especially, when thinking about the derived numbers

- Risk Balance: Simply the threshold the balance cannot go above.
- An Agent Transaction (energy or appliance sale) adds to this number.
- Transfer to company decreases the number.
- Commission sales decreases this number.

A hypothetical "company gives money to the agent" would also add to this number.

WDYT???


Now, to achieve this a few steps have to be taken.

- A migration that `* (-1)` all entries in AgentHistory
- Update all the validation logic to consider the updated meaning.

Note: Special care has to be taken for AgentCharge because it super broken right now. According to the docs AgentCharge should represent if the company gives/borrows money to the agent. In the old terminology this should result in a negative entry: "The Agent now owes the company". But in fact, it's using the intuitive signs:

image

beesaferoot: > Thanks for looking into that 🙏 Unfortunately (for me) it doesn't change much. We are "just" adding new terminology.

I understand this,  I initially wanted to defer this as it may cause delay to actually getting issues that users are facing from getting out early. On second thought, I see how it will serve them better if we simplify things rather than explain it away. 

We may have some differences to how the "Agent Charge" should work but all other areas have been worked.

docs/usage-guide/agent-users.md
```diff
@@ -18,13 +18,27 @@ For more information on how to generate and manage tickets, see [Tickets](/usage
...
 Agents receive cash from customers on site.
-An agent's **outstanding balance** is the money it owes the company (i.e. it is the money it has collected on behalf of the company and has to transfer accordingly).
-
-The **risk balance** is the maximum amount of money that an agent can collect before it has to transfer the money to the company.
-When that balance is reached, agents will not be able to collect money anymore (generate tokens) or sell appliances (his account on the app will not work.
-So, when the agent sends the balance money to the company, then headquarters has to manually adjust the outstanding balance by that agent through MicroPowerManager desktop.
+MicroPowerManager tracks two totals for every agent, both visible on the agent's profile card:
+
+- **Balance**: the amount of company money the agent currently holds.
+  It is a single, positive figure.
```
dmohns: You can remove `It is a single, positive figure.`, it only makes sense in light of the change. For a new user this is what they would expect and only causes the confusion: "Why is this mentioned here?"

docs/usage-guide/agent-users.md
```diff
@@ -18,13 +18,27 @@ For more information on how to generate and manage tickets, see [Tickets](/usage
...
+So, when the agent hands the collected money to the company, headquarters records this through MicroPowerManager desktop.
 This is done by creating a **receipt** on the specific agent’s profile.
-When that receipt is created, balance of agent is removed/decreased to 0.
+When that receipt is created, the handed-over cash comes off the balance and the agent's earned commission is credited to them as an explicit **Payout** entry, which also reduces the balance.
+
+::: warning Balance sign change after the upgrade
+For existing installations, the agent balance figures flip sign after this upgrade.
+What used to display as a negative balance while the agent held company cash now displays as a positive "money held" figure.
+The number is the same amount — only the sign and its meaning changed — so there is no cause for alarm.
+:::
```
dmohns: This needs to go into the Release Notes, not in the usage guide. 

I'd be fine to keep a small note here. But then we'd need to change the wording. Some thing along the lines of

"Prior to version (X.Y.Z) agent balance were displayed as negative...."

dmohns (dismissed):
This is such a great improvement. Thanks for the effort 🙏 

I have a few nits regarding documentation. I'll approve the PR already, you can address the docs stuff in a separate PR if you want.

docs/usage-guide/agent-users.md
beesaferoot: Noted.

docs/usage-guide/agent-users.md
```diff
@@ -18,13 +18,25 @@ For more information on how to generate and manage tickets, see [Tickets](/usage
...
+When a sale would push the balance above the risk balance, agents will not be able to collect money anymore (generate tokens) or sell appliances (their account on the app will not work).
+So, when the agent hands the collected money to the company, headquarters records this through MicroPowerManager desktop.
 This is done by creating a **receipt** on the specific agent’s profile.
-When that receipt is created, balance of agent is removed/decreased to 0.
+When that receipt is created, the handed-over cash comes off the balance and the agent's earned commission is credited to them as an explicit **Payout** entry, which also reduces the balance.
+
+::: tip Note on the balance sign
+Prior to version 0.1.34, an agent holding company cash was displayed as a negative balance.
+It is now shown as a positive "money held" figure of the same amount — only the sign and its meaning changed.
+:::
```
dmohns: Another small nit: Vitepress support both [Github style](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) `> [!TIP]` and [native Admonition Containers](https://vitepress-theme-default-plus.lando.dev/markdown/admonitions.html) `::: tip`

Currently we use Github style everywhere as it renders nicely if the files is looked at through Github. For example: https://github.com/EnAccess/micropowermanager/blob/main/docs/development/development-environment.md#reseting-the-demo-data

beesaferoot: Updated.

dmohns (approved):

PR #1577, opened 2026-07-22
feat: add amount variable/value in shs token confirmation sms message

### LINKED ISSUE #1561: [Feature Request]: SMS for Token confirmation should also include transaction amount.
### Preflight Checklist

### Problem Description

Feature request from user that wants to include amount for the token confirmation SMS. Sometimes customers request to see the amount information. 

### Proposed Solution

Extract the amount information from the transaction object and provide the amount variable for SHS token confirmation template

### Alternatives Considered

N/A

### Additional Information

_No response_

beesaferoot:
Closes #1561 

### Pull Request checklist


src/backend/database/migrations/tenant/2026_07_22_120000_add_amount_to_token_confirmation_shs_sms_body.php
```diff
@@ -0,0 +1,32 @@
...
+    public function up(): void {
+        DB::connection('tenant')->table('sms_bodies')
+            ->where('reference', 'TokenConfirmationSHS')
+            ->update([
+                'body' => self::BODY_WITH_AMOUNT,
+                'place_holder' => self::BODY_WITH_AMOUNT,
+                'variables' => 'name,surname,token,duration,unit,device_serial,amount',
+                'updated_at' => Carbon::now(),
+            ]);
+    }
```
dmohns, line 35: I don't think we can do this. This will overwrite (and delete) any customisation a user has done. 

You would have to do (in pseudo):

IF `body ==  self::BODY_WITHOUT_AMOUNT` THEN UPDATE ELSE NOTHING

btw... what's the difference between `body` and `place_holder`?

beesaferoot: Thanks for spotting this. I think we I wanted to update is the place_holder 

> btw... what's the difference between body and place_holder?

The body is what can be customized by the user while the place_holder is what we initially set and is not be updated by the user.

dmohns: Ohhh, I see... What an over-complicated, unintuitive behaviour, haha 🙈

src/backend/database/migrations/tenant/2026_07_22_120000_add_amount_to_token_confirmation_shs_sms_body.php
```diff
@@ -0,0 +1,30 @@
...
+    private const BODY_WITHOUT_AMOUNT = 'Dear [name] [surname], your transaction is confirmed. Device Serial: [device_serial]. Token: [token], Duration: [duration] [unit].';
+
+    public function up(): void {
+        DB::connection('tenant')->table('sms_bodies')
+            ->where('reference', 'TokenConfirmationSHS')
+            ->update([
+                'place_holder' => self::BODY_WITH_AMOUNT,
+                'variables' => 'name,surname,token,duration,unit,device_serial,amount',
+                'updated_at' => Carbon::now(),
+            ]);
```
dmohns, line 34: It still doesn't work. On new installations/tenants it will have the `body` from before, i.e. without amount.

dmohns:
Sorry, my comments are all over this PR 🙈 

From what I understand `place_holder` is essentially a useless field. It doesn't do anything, right? I _think_ what it was meant to be is `default_value`. This would make some sense if we had a button to "Reset SMS Bodies to default" (which we don't). But this is probably a future improvement.

src/backend/database/migrations/tenant/2026_07_22_120000_add_amount_to_token_confirmation_shs_sms_body.php
dmohns: I think for now, we have to stick to the bahaviour:

- Always update `place_holder`
- Update `body` if it was untouched else keep the user-modified body

beesaferoot:
> From what I understand `place_holder` is essentially a useless field. It doesn't do anything, right? I _think_ what it was meant to be is `default_value`. This would make some sense if we had a button to "Reset SMS Bodies to default" (which we don't).

One missing part to this is that the placeholder value is used when the user clears the textarea as a hint, so I think it is still useful. 

Screenshot 2026-07-27 at 10 48 42 AM

dmohns (approved):

PR #1621, opened 2026-08-12
feat(payment-providers): Add support to close registration tail wizard

### LINKED ISSUE #1620: [Feature Request]: Registration Tail wizard modal doesn't close when a step navigates away to configure
### Preflight Checklist

### Problem Description

When a plugin is enabled but not yet configured, the post-login Registration Tail wizard shows a modal dialog with a "Configure {Plugin}" button. Clicking it navigates to the plugin's Overview page but the wizard modal stays open, overlaying the very page the user was just sent to configure.

Steps to reproduce :
1. Enable a payment plugin without configuring credentials yet, and log in (or log out and back in) so the Registration Tail wizard appears.
2. Click "Configure {Plugin}" on the wizard card.

Expected: the wizard dialog closes and the user lands cleanly on the plugin's Overview page.
Actual: the app navigates correctly, but the dialog remains open on top of the page.

A related, secondary bug surfaced from the same flow: if the user was already sitting on the target route (e.g. /flutterwave/overview) when clicking "Configure," the button's `this.$router.push(...)` call rejected with an uncaught NavigationDuplicated error (vue-router 3.x throws when navigating to the current route), producing a visible "Uncaught runtime errors" dev overlay. That part is fixed by catching and ignoring NavigationDuplicated on the push, same as the existing pattern in shared/Paginate.vue.

Root cause

The wizard dialog's open/closed state (wizardIsVisible) is owned entirely by shared/TailWizard.vue. The plugin-specific card (e.g. plugins/flutterwave-payment-provider/Component.vue) is rendered as a disconnected dynamic child inside that dialog (<component :is="tailObj.component" />) and has no reference back to it, so its "Configure" button, which only calls `this.$router.push(...)`, has no way to tell the dialog to close.

The only existing signal TailWizard.vue listens for (EventBus.$on(tailObj.component, ...)) both closes the dialog and marks the registration-tail step as adjusted: true on the backend however using it here would be wrong, since the user hasn't actually saved any credentials yet; it would prematurely mark the step "done."

### Proposed Solution

Added a second, purpose-built signal that only closes the dialog without touching the backend adjusted state:

- shared/TailWizard.vue: listens for a new tail-wizard.close event on the shared EventBus, calling a new closeWizard() method that dismisses the dialog exactly the way the existing "Do this later" button does (wizardIsVisible = false, marks isWizardShown in the Vuex store so it won't reopen for the rest of the session).
- plugins/flutterwave-payment-provider/Component.vue: emits tail-wizard.close right before navigating in goToConfiguration().

This change is purely additive to `TailWizard.vue `- no other plugin emits the new event, so Paystack, Pesapal, and every other plugin using this wizard keep their current behavior unchanged. The fix was intentionally scoped to Flutterwave only.

### Alternatives Considered

 N/A

### Additional Information

_No response_

Dijeomah:
Fixes the Registration Tail wizard modal for Paystack and Pesapal: clicking "Configure {Plugin}" on the wizard card navigated to the plugin's Overview page but left the modal dialog open on top of it, and threw an uncaught NavigationDuplicated error if the user was already on that route.

- shared/TailWizard.vue: added a new tail-wizard.close event on the shared EventBus, handled by a new closeWizard() method that dismisses the dialog the same way the existing "Do this later" button does without marking the registration-tail step as adjusted.
- both vue components (plugins/paystack-payment-provider/Component.vue and plugins/pesapal-payment-provider/Component.vue): goToConfiguration() now emits tail-wizard.close before navigating.

closes: #1620

None expected,  it doesn't alter the existing "Do this later" or step-completion behavior for any plugin.

- (Manual)  recommend before merge:
  a. Enable Paystack (or Pesapal) without configuring credentials, log out and back in so the wizard appears.
  b. Click "Configure Paystack"/"Configure Pesapal" - confirm the modal closes and the Overview page is reachable underneath.

### Pull Request checklist

- [X] Meaningful Pull Request title and description
- [X] Changes tested as described above


src/frontend/src/plugins/paystack-payment-provider/Component.vue
```diff
@@ -77,7 +78,14 @@ export default {
   },
   methods: {
     goToConfiguration() {
-      this.$router.push("/paystack/overview")
+      // Only closes the dialog — the registration-tail step must stay unadjusted
+      // until credentials are actually saved on the overview page.
+      EventBus.$emit("tail-wizard.close")
```
beesaferoot: I think we can consolidate this action in one place and reuse it in other plugins.

src/frontend/src/plugins/pesapal-payment-provider/Component.vue
```diff
@@ -77,7 +78,14 @@ export default {
...
     goToConfiguration() {
-      this.$router.push("/pesapal/overview")
+      // Only closes the dialog — the registration-tail step must stay unadjusted
+      // until credentials are actually saved on the overview page.
+      EventBus.$emit("tail-wizard.close")
+      this.$router.push("/pesapal/overview").catch((error) => {
+        if (error.name !== "NavigationDuplicated") {
+          throw error
+        }
+      })
```
beesaferoot: Mentioned above

beesaferoot (commented):
Hi @Dijeomah, thanks for spotting and creating fixes for this.

Added a small comment

src/frontend/src/plugins/pesapal-payment-provider/Component.vue
Dijeomah: Can you confirm, there is an update to this branch.

beesaferoot (approved):

